PR #1, opened 2015-02-27
Discussion: Strongly-type entrypoint arguments

kofalt:
As in hroot, we may as well provide an array for a command and its arguments.


heavenlyhash:
LGTM

PR #2, opened 2015-03-02
Schema discussion

kofalt:
Simple changes:

* Combined OutputSpec into Output. Separation unnecessary. 
* Rename ActiveJob -> Job, rename JobRecord -> JobDraft. I feel like this more accurately represents their purpose; a JobDraft is a set of instructions to create a Job (possibly multiple times), a Job may finish and still exist as a concept. For example Job.Start, Job.Wait.


heavenlyhash:
Yeah, getting rid of `OutputSpec` as a distinction and just using a zero-value `Hash` will make things easier.  We're actually going to want to hash the whole set of outputs in defining the first half of the "conjecture" as well (since having the same sum of outputs on a different configuration of outputs would be silly), and having just one hashing method for that structure will save code.

def/def.go
```diff
@@ -24,20 +24,17 @@ type Accents struct {
...
 	Conjecture bool   // If an `Output` is part of a job's 'conjecture', it is expected to produce the same result, every time, when given the same set of `Input` items.  This can be used to verify correct repeatability of a process.  An example of using (and not using) this: the set of which tests pass and fail should be consistent for a given input git repo, so output lists of pass/fail test names in an output that is conjecture=true; meanwhile, the logs from the tests probably contain timestamps and so aren't part of precise repeatability, so while we want to keep those for reading, they belong in another output that's conjecture=false.
-}
 
-type Output struct {
-	OutputSpec
 	Hash string // `Output.Hash` is generated by the output handling implementation.  (In a content-addressible filesystem, we can just borrow their concept.  For other more legacy-oriented systems, this may be a hash of the contents produced from the job's working filesystem before export.)
```
heavenlyhash: Move `Hash` up to above `URL` to match order in `Input` struct?

kofalt:
Agreed.

def/def.go
kofalt: Sure :P

heavenlyhash:
How about "formula" for the struct rename?  Nicely captures the "A + B -> C" nature of where we're going with this.  And a single noun will be nice since we're going to refer to this a *lot* in the course of documentation.

kofalt:
:+1:

heavenlyhash:
LGTM

PR #3, opened 2015-03-02
Dreams of CLI

kofalt:
A basic CLI entry point for repeatr.

Can specify an executor and a (currently must by JSON) input file to use as forumla.

A basic stub that can grow into a lovely flower!

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,0 +1,47 @@
+package cli
+
+import (
+	. "fmt"
+	"os"
+
+	"github.com/codegangsta/cli"
+
+	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/executor"
+	"polydawn.net/repeatr/executor/null"
+)
+
+func GetApp() *cli.App {
+	app := cli.NewApp()
+
+	app.Name = "repeatr"
+	app.Usage = "Run it. Run it again."
+	app.Version = "0.0.1"
+
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "executor, e",
+			Value: "null",
+			Usage: "Which executor to use",
+		},
+	}
+
+	app.Action = func(c *cli.Context) {
+
+		desiredExecutor := c.String("executor")
+
+		var executor executor.Executor
+
+		switch desiredExecutor {
+		case "null":
+			executor = &null.Executor{}
+		default:
+			Println("No such executor", desiredExecutor)
+			os.Exit(1)
+		}
+
+		executor.Run(def.Formula{})
+	}
+
+	return app
+}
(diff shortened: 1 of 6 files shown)


cli/cli.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/codegangsta/cli"
+	"github.com/ugorji/go/codec"
+
+	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/executor"
+	"polydawn.net/repeatr/executor/null"
+)
+
+func GetApp() *cli.App {
+	app := cli.NewApp()
```
heavenlyhash: This could just be a package var and the rest in an `init()` function.  We'll never want more than one of them.

cli/cli.go
```diff
@@ -0,0 +1,77 @@
...
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "executor, e",
+					Value: "null",
+					Usage: "Which executor to use",
+				},
+				cli.StringFlag{
+					Name:  "input, i",
+					Value: "formula.json",
+					Usage: "Location of input forumla (json format)",
```
heavenlyhash: splchk

cli/cli.go
```diff
@@ -0,0 +1,77 @@
...
+			Usage:  "Run a forumla",
+			Action: Run,
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "executor, e",
+					Value: "null",
+					Usage: "Which executor to use",
+				},
+				cli.StringFlag{
+					Name:  "input, i",
```
heavenlyhash: We're already overloading the word 'input' in way too many places, what do you think of calling this 'formula' in the flag instead?

cli/cli.go
```diff
@@ -0,0 +1,77 @@
...
+	content, err := ioutil.ReadFile(filename)
+	if err != nil {
+		Println(err)
+		Println("Could not read file", filename)
+		os.Exit(1)
+	}
+
+	var js codec.JsonHandle
+	handler := &js
+	dec := codec.NewDecoderBytes(content, handler)
```
heavenlyhash: can just terse this up to `dec := codec.NewDecoderBytes(content, &codec.JsonHandle{})`

cli/cli.go
kofalt: Eh... We'll want to be able to invoke many times for integration tests, perhaps. But the various cli `Run()` functions do not appear to mutate the struct, so I'll use `init()` because why not :)

cli/cli.go
kofalt: I thought about it, but `-f` feels really dumb to me.

cli/cli.go
kofalt: Meh. Was blindly copying example code, changed

cli/cli.go
kofalt: Changed

kofalt:
Rebased onto master to fix merge conflict.

heavenlyhash:
LGTM

PR #4, opened 2015-03-02
Add dependency on spacemonkey errors.

heavenlyhash:
Or rather, our fork of it.  This one includes handlers and helpers for panic-based flow control.

(not much to see here, but a lot of future work will use this.)


kofalt:
LGTM

PR #5, opened 2015-03-04
Inputs draft

heavenlyhash:
First pass on some hashing and file shuffling stuff for inputs.

There's still a lot of todos here, but any tests not yet passing are skipped so we can merge this and continue to iterate from here.

### DIFF
--- a/input/dir/dir_input.go
+++ b/input/dir/dir_input.go
@@ -1,0 +1,55 @@
+package dir
+
+import (
+	"crypto/sha512"
+	"hash"
+	"os"
+
+	"github.com/spacemonkeygo/errors"
+	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/input"
+	"polydawn.net/repeatr/lib/fshash"
+)
+
+const Type = "dir"
+
+// interface assertion
+var _ input.Input = &Input{}
+
+type Input struct {
+	spec          def.Input
+	hasherFactory func() hash.Hash
+}
+
+func New(spec def.Input) *Input {
+	if spec.Type != Type {
+		panic(errors.ProgrammerError.New("This input implementation supports definitions of type %q, not %q", Type, spec.Type))
+	}
+	_, err := os.Stat(spec.URL)
+	if os.IsNotExist(err) {
+		panic(def.ValidationError.New("Input URL %q must be a directory", spec.URL))
+	}
+	return &Input{
+		spec:          spec,
+		hasherFactory: sha512.New384,
+	}
+}
+
+func (i Input) Apply(destinationRoot string) <-chan error {
+	done := make(chan error)
+	go func() {
+		defer close(done)
+
+		// walk filesystem, copying and accumulating data for integrity check
+		bucket := &fshash.MemoryBucket{}
+		err := fshash.FillBucket(i.spec.URL, destinationRoot, bucket, i.hasherFactory)
+		if err != nil {
+			done <- err
+			return
+		}
+
+		// hash whole tree and verify total integrity
+		// TODO
+	}()
+	return done
+}
--- a/input/dir/dir_input_test.go
+++ b/input/dir/dir_input_test.go
@@ -5,11 +5,13 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	. "github.com/smartystreets/goconvey/convey"
 	"github.com/spacemonkeygo/errors"
 	"github.com/spacemonkeygo/errors/try"
 	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/lib/fshash"
 	"polydawn.net/repeatr/testutil"
 )
 
@@ -53,6 +55,12 @@ func Test(t *testing.T) {
 				})
 
 				Convey("Apply succeeds (hash checks pass)", func() {
+					// wait a moment before copying to decrease the odds of nanotime telling a huge, huge lie.
+					// this, incidentally, varies in effectiveness with whether your running with the goconvey web app or not.
+					// yes, i'm serious.  i clicked many times to determine this.
+					// this is literally the reason why we're building repeatr.
+					time.Sleep(2 * time.Millisecond)
+
 					waitCh := inputter.Apply(filepath.Join(pwd, "dest"))
 					So(<-waitCh, ShouldBeNil)
 
@@ -65,7 +73,19 @@ func Test(t *testing.T) {
 						So(string(content), ShouldEqual, "zyx")
 
 						Convey("And all metadata matches", func() {
-
+							// Comparing fileinfo doesn't work conveniently; you keep getting new pointers for 'sys'
+							//one, _ := os.Lstat("src/a")
+							//two, _ := os.Lstat("dest/a")
+							//So(one, ShouldResemble, two)
+							So(fshash.ReadMetadata("src/a"), ShouldResemble, fshash.ReadMetadata("dest/a"))
+							So(fshash.ReadMetadata("src/b"), ShouldResemble, fshash.ReadMetadata("dest/b"))
+							So(fshash.ReadMetadata("src/b/c"), ShouldResemble, fshash.ReadMetadata("dest/b/c"))
+							// the top dir should have the same attribs too!  but we have to fix the name.
+							srcDirMeta := fshash.ReadMetadata("src/")
+							srcDirMeta.Name = ""
+							destDirMeta := fshash.ReadMetadata("dest/")
+							destDirMeta.Name = ""
+							So(srcDirMeta, ShouldResemble, destDirMeta)
 						})
 					})
 				})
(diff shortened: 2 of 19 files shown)


.gopath/src/github.com/jtolds/gls
```diff
@@ -0,0 +1 @@
+Subproject commit f1ac7f4f24f50328e6bc838ca4437d1612a0243c
```
kofalt: wat

Please explain?

input/dir/dir_input.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/spacemonkeygo/errors"
+	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/input"
+	"polydawn.net/repeatr/lib/fshash"
+)
+
+const Type = "dir"
+
+// interface assertion
+var _ input.Input = &Input{}
```
kofalt: Thinking out loud: maybe we want these in tests? 

Alternatively, I'm about to push a `dispatch` package that accomplish this same end elsewhere. Maybe I'll just remove these assertions in my branch.

input/dir/dir_input.go
```diff
@@ -0,0 +1,55 @@
...
+
+type Input struct {
+	spec          def.Input
+	hasherFactory func() hash.Hash
+}
+
+func New(spec def.Input) *Input {
+	if spec.Type != Type {
+		panic(errors.ProgrammerError.New("This input implementation supports definitions of type %q, not %q", Type, spec.Type))
```
kofalt: For similar reasons to my previous comment, if we have a dispatch method this check will never be necessary. I feel more comfortable having that enforced in a single location rather than in each implementation.

input/dir/dir_input.go
heavenlyhash: I like keeping them near every typedef.  It's the least confusing place to get the error emitted from.  (The Go style guide disagrees that you should even use this trick; IMO that's sensible if you're composing detachable packages, but utter nonsense if the interface is guaranteed to be in your imported packages anyway and it's critical to sanity.)

.gopath/src/github.com/jtolds/gls
heavenlyhash: Transitive dep of goconvey since we last used it -.-

input/dir/dir_input_test.go
```diff
@@ -0,0 +1,103 @@
...
+		Convey("The input config should be rejected during validation", func() {
+			correctError := false
+			try.Do(func() {
+				New(def.Input{
+					Type:     "dir",
+					Hash:     "abcd",
+					URL:      "/tmp/certainly/should/not/exist",
+					Location: "/data",
+					// REVIEW: two of these four fields were more for the executor.
+					// maybe this struct isn't the sanest to use for this arg.
```
kofalt: I agree that giving the Input impl too much information isn't desirable. How would you feel about a smaller struct that is embedded by Input?

input/dir/dir_input_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			So(correctError, ShouldBeTrue)
+		})
+	})
+
+	Convey("Given a directory with a mixture of files and folders", t,
+		testutil.WithTmpdir(func() {
+			pwd, _ := os.Getwd()
+			os.Mkdir("src", 0755)
+			os.Mkdir("src/a", 01777)
```
kofalt: Is this setting the sticky bit on `src/a`? Refresh my memory on linux fs -.-

input/dir/dir_input_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			Convey("We can construct an input", func() {
+				inputter := New(def.Input{
+					Type: "dir",
+					Hash: "abcd",
+					URL:  filepath.Join(pwd, "src"),
+				})
+
+				Convey("Apply succeeds (hash checks pass)", func() {
+					// wait a moment before copying to decrease the odds of nanotime telling a huge, huge lie.
```
kofalt: Can you amend this slightly to explain *what* goes wrong, exactly? There's no context.

lib/fspatch/NOTICE
```diff
@@ -0,0 +1,11 @@
...
+  https://github.com/docker/docker/tree/master/pkg/system
+
+The upstream license is also Apache2, and we continue our modifications
+under the same license.
+
+We happily credit the origin of this code to those fine authors.  We regret
+any inconvenience in historical tracking; we'd prefer to maintain a proper
+hashed linkage to the original repository and reuse the code as a library
+directly, but ~100 megs (compressed) of vendored unrelated deps in that repo
+makes that untenable.    ヽ(´ー｀)ノ
```
kofalt: rumble

testutil/assertions.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+		modeType := info.Mode() & os.ModeType
+		if modeType != mode {
+			return fmt.Sprintf("Expected file to have mode %v but it had %v instead!", mode, modeType)
+		}
+		return ""
+	default:
+		return "You must provide zero or one parameters as expectations to this assertion."
+	}
+}
```
kofalt: Could this be a PR to github.com/smartystreets/assertions ?

input/dir/dir_input.go
kofalt: I'm mainly thinking ahead to whenever we might refactor an interface and be unable to build until we adapt all implementations. *shrug*

Let's leave it for now, can always remove later.

testutil/assertions.go
heavenlyhash: I don't think they need a PR for every crappy assertion this ape with two thumbs spits out and has no intention of generalizing.  They document a way to do extensions without upstreaming them for a reason.

input/dir/dir_input_test.go
heavenlyhash: Yup.  I just picked an amusing set of numbers; all of which are covered by the assertions that load a tar header, since that includes mode bits.

input/dir/dir_input_test.go
heavenlyhash: Nanotime is *always* a lie.  Your OS can and will randomly decide to change its time reporting granularity according to its ambient temperature and how busy it feels -- even up to the order of milliseconds.  Even under "high"(ish) granularity states (read: my box at rest before I started the test), this test can totally go fast enough that all of the times writing the test files and all of the times on the files created by the copy just happen to be in the "same" nanosecond... like, yes, without this pause, the test will pass *even without* the code for forcing the fs timestamps on the copied files.

input/dir/dir_input_test.go
heavenlyhash: /shrug, maybe later?  Reconsidering it now by the light of day, it seems kinda silly to worry about too many fields being used in a code path that's basically only ever going to be hit directly by tests.  I'll remove the comment for now and we can reconsider if it still seems awkward in the bigger picture later.

input/dir/dir_input.go
heavenlyhash: Think of it like the panic in the `default` of a switch handling an enum.  Actually hitting this would be a compile-time fuckup, yes, but I for one make mistakes like that all the time, and that's also why it's in an error hierarchy that's never supposed to be caught.

lib/fspatch/NOTICE
heavenlyhash: Don't worry!  The entire core team is about to go [full vendor](https://groups.google.com/forum/#!topic/golang-dev/nMWoEAG55v8)!

Everybody knows you never do a full vendor.

But seriously, thanks to Alexander Larsson and Victor Vieux who afaict are the origins of this solution.

I really wish there was a better way to handle this than copyforking.

heavenlyhash:
Comments addressed and a few other fixes applied, should be good now.

.gopath/src/github.com/jtolds/gls
kofalt: Oof. Reluctantly accepted :(

input/dir/dir_input.go
kofalt: Still feels like belt & suspenders to me, but it may as well stay for now. Conversation will likely make more sense in light of a executor PR that I'll file later today.

lib/fspatch/NOTICE
kofalt: In the future we can maybe handle this with polydawn-ports autofilter. If that ever ends up existing, we'll be able to link against `github.com/polydawn-ports/docker-lean` or something

input/dir/dir_input_test.go
kofalt: The **runtime variance** on reporting accuracy is new information to me! Very interesting. I see you added an addition clarifying that; good.

testutil/assertions.go
kofalt: Poor apes.

kofalt:
LGTM!

lib/fspatch/NOTICE
heavenlyhash: Yeah probably the most relieving thing on that golang-dev thread is the suggestion that by standardizing vendoring locations and metadata, it'll be (more conveniently) possible to mechanically strip unwanted stuff back out as a third party. It'll still create anti-historic psuedoforking wasteoftime nonsense, but...

PR #7, opened 2015-03-05
Executor prep

kofalt:
These are all prerequisites of the upcoming draft nsinit executor. Makes sense to review them here.

* Dispatch methods to simplify finding impls (stub-ish thing, can be overhauled)
* Validator to perform common formula sanity checking + massaging
* The inevitable util-ish package, currently with a single convenience method for scoped temp dirs

Spacemonkey errors just landed, so I'll update this PR with some strongly-typed errors.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -1,0 +1,60 @@
+package dispatch
+
+import (
+	. "fmt"
+	"os"
+
+	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/executor"
+	"polydawn.net/repeatr/executor/null"
+	"polydawn.net/repeatr/input"
+	"polydawn.net/repeatr/input/dir"
+	"polydawn.net/repeatr/input/tar"
+	"polydawn.net/repeatr/output"
+)
+
+// TODO: This should not require a global string -> class map :|
+// Should attempt to reflect-find, trying main package name first.
+// Will make simpler to use extended transports, etc.
+
+func GetExecutor(desire string) *executor.Executor {
+	var executor executor.Executor
+
+	switch desire {
+		case "null":
+			executor = &null.Executor{}
+		default:
+			Println("No such executor", desire)
+			os.Exit(1)
+	}
+
+	return &executor
+}
+
+func GetInput(desire def.Input) *input.Input {
+	var input input.Input
+
+	switch desire.Type {
+		case "dir":
+			input = dir.New(desire)
+		case "tar":
+			input = tar.New(desire)
+		default:
+			Println("No such input", desire)
+			os.Exit(1)
+	}
+
+	return &input
+}
+
+func GetOutput(desire def.Output) *output.Output {
+	var output output.Output
+
+	switch desire.Type {
+		default:
+			Println("No such output", desire)
+			os.Exit(1)
+	}
+
+	return &output
+}
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -21,11 +21,11 @@ func GetExecutor(desire string) *executor.Executor {
 	var executor executor.Executor
 
 	switch desire {
-		case "null":
-			executor = &null.Executor{}
-		default:
-			Println("No such executor", desire)
-			os.Exit(1)
+	case "null":
+		executor = &null.Executor{}
+	default:
+		Println("No such executor", desire)
+		os.Exit(1)
 	}
 
 	return &executor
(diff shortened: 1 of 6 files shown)


lib/flak/assist.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Generates a temporary repeatr directory, creating all neccesary parent folders.
+// Must be passed at least one directory name, all of which will be used in the path.
+// Uses os.TempDir() to decide where to place.
+//
+// For example, GetTempDir("my-executor") -> /tmp/repeatr/my-executor/989443394
+func GetTempDir(dirs ...string) string {
+
+	if len(dirs) < 1 {
+		Println("Must have at least one sub-folder for tempdir", "replace with space monkey")
```
heavenlyhash: other merges have gone through, you have the power.

i'd use `ProgrammerError` since this is a compile-time fuckup (`ValidationError` is the other potential sensible category, but those are more for when user input processing is involved and thus can make sense to catch in normal flow).

lib/flak/assist.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	dir := []string{os.TempDir(), "repeatr"}
+	dir = append(dir, dirs...)
+	tempPath := filepath.Join(dir...)
+
+	// Tempdir wants parent path to exist
+	err := os.MkdirAll(tempPath, 0600)
+	if err != nil {
+		Println(err, "replace with space monkey")
```
heavenlyhash: `errors.IOError.Wrap(err)`

dispatch/dispatch.go
```diff
@@ -0,0 +1,60 @@
...
+	"polydawn.net/repeatr/executor/null"
+	"polydawn.net/repeatr/input"
+	"polydawn.net/repeatr/input/dir"
+	"polydawn.net/repeatr/input/tar"
+	"polydawn.net/repeatr/output"
+)
+
+// TODO: This should not require a global string -> class map :|
+// Should attempt to reflect-find, trying main package name first.
+// Will make simpler to use extended transports, etc.
```
heavenlyhash: /shrug.  I'm fine with maintaining this as a string switch somewhere.  It's the path of least resistance and doesn't box anything in.  We might want characters that don't fit in a package name later; we might end up with some versioning scheme in the name that requires parsing; etc.  I've been following a rough convention of having a `$pkg.Type string` in every input package and moving on.  Another alternative is having a big globally visible map and having implementations register themselves in the package init function, but then you still have to *touch* the package, which means your import list just turned into a bunch of "_" imports and it's a bunch of strings you have to maintain again anyway.

kofalt: I'm thinking ahead here to outside-this-repository implementations. 

We can totally maintain a string mapping until the end of time, and you're right in that package strings might not always be a ceiling here, but I'd like to be able to put

```json
{
	"Type": "not.your.domain.com/lol/wat",
	"Location": "/var/log",
	"URL": "whelp.tar"
}
``` 

And have that Just Work. 

Or to clarify further, scitran engine will be rolling our own scheduler and almost certainly some custom transports; importing most of repeatr and interoperating it would be nice.

A reflect lookup could be added as a final lookup, post-strings.

lib/flak/assist.go
kofalt: Works for me. I'll be adding a commit with some created error types; discussion desired.

lib/flak/assist.go
kofalt: Works.jpg. I take it we are accepting these as cannon?

http://godoc.org/github.com/spacemonkeygo/errors#pkg-variables

heavenlyhash: yeah, I'm rolling with them until we have overwhelming reason not to.

panic/return to taste.  Sometimes (i.e. the file walker, where it's convenient to do so) I'm still using error returns, but most things that have a complaint to make are more than one functional call deep and for those, panics are fine.  When panicing, then use http://godoc.org/github.com/polydawn/errors/try to control.

kofalt: SGTM; will do!

kofalt:
Now that other PRs have landed, I added spacemonkey errors usage & rebased.

dispatch/dispatch.go
```diff
@@ -0,0 +1,56 @@
...
+
+func GetExecutor(desire string) *executor.Executor {
+	var executor executor.Executor
+
+	switch desire {
+	case "null":
+		executor = &null.Executor{}
+	default:
+		panic(errors.NotImplementedError.New("No such executor %s", desire))
```
heavenlyhash: If returning nil doesn't seem fun here, I'd consider these more `ValidationError`: this function is presumably handling user input or config data since otherwise the caller would just use the reference they meant directly, and this error is reached when the caller asked for a name that doesn't make sense.  NotImplemented would make more sense if it was like `case "coolPlannedThing": panic(NotImplemented)`.

heavenlyhash:
woot for error handling and ci green; full steam ahead :steam_locomotive:

PR #8, opened 2015-03-05
Treewalker

heavenlyhash:
A simple iterator interface that, when implemented, lets you conveniently do a visitor pattern walk on a tree.  Critically, with both pre-order and post-order visitation... which means after this is merged, I can replace `filepath.Walk` and fix the remaining skipped tests in the `dir` input implementation that need post-order visitation to build.  Right after that, also will be useful for walking trees in the hash buckets and building the final hash.

Consumer code tends to end up with a type assertion as the first step in every `WalkFunc`.  I can't think of any way to make this suck less... of course Generics would, but... golang.

(More general than repeatr?  Absolutely.  Resisting the temptation to put it in a grabbag "golang-utils" library until we actually discover an impetus to do so.)

### DIFF
--- a/lib/treewalk/walk.go
+++ b/lib/treewalk/walk.go
@@ -8,10 +8,6 @@ type Node interface {
 	NextChild() Node
 }
 
-func Walk(root Node, preVisit WalkFunc, postVisit WalkFunc) error {
-	return walk(root, preVisit, postVisit)
-}
-
 type WalkFunc func(node Node) error
 
 /*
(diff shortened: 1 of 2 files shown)


lib/treewalk/walk.go
```diff
@@ -0,0 +1,57 @@
...
+	"errors"
+)
+
+type Node interface {
+	NextChild() Node
+}
+
+func Walk(root Node, preVisit WalkFunc, postVisit WalkFunc) error {
+	return walk(root, preVisit, postVisit)
+}
```
kofalt: Why the indirection here?

heavenlyhash: Wow that's... a really good question.  I think because this evolved by mutation from `filepath.Walk` and that had a bunch of internal complications that made starting different than the rest of traversal, and I got rid of that in this implementation.  Removing.

PR #11, opened 2015-03-09
Hashing the bucket

heavenlyhash:
Headline: The `dir` input system is now essentially functional!

Full metadata, deterministic serialization, tree hash over the whole set.  Any potentially meaningful changes result in a change in the hash.

This'll also be almost entirely reused by the tar input types momentarily, and of course all the hashing systems will also be reusable for any outputs that don't have their own built-in content-addressability.

We'll probably want to strap the fshash functions for generating this into some sort of cli "scan" command for getting the `Hash` of data that's being entered to the system for the first time, but details of that can be hashed out in the future.

(Diff is large and there's at least one reorg commit buried in here so it may be easier to review commitwise via `git log -p`.)


heavenlyhash:
Ironically, this fails in travis because UIDs and GIDs are included here, and I have different ones locally than they do, so the one hash fixture assertion fails.  We'll have to do filters sooner than later.

lib/fshash/bucket.go
```diff
@@ -58,9 +15,30 @@ func (m Metadata) MarshalBinary() ([]byte, error) {
 */
 type Bucket interface {
 	Record(metadata Metadata, contentHash []byte)
+	Iterator() (rootRecord RecordIterator)
+}
+
+type Record struct {
+	Metadata    Metadata `json:"m"`
+	ContentHash []byte   `json:"h"`
```
kofalt: Is this an example of codec using the json annotation?

lib/fshash/metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	"path/filepath"
+	"sort"
+	"time"
+
+	"github.com/spacemonkeygo/errors"
+	"github.com/ugorji/go/codec"
+	"polydawn.net/repeatr/def"
+)
+
+type Metadata tar.Header
```
kofalt: On the assumption that we're neither overly fond of nor desire to be tied to tarballs, is it worth considering if we want to whitelist our own fields? OTOH, it is pretty close to what we want:

```go
type Header struct {
        Name       string    // name of header file entry
        Mode       int64     // permission and mode bits
        Uid        int       // user id of owner
        Gid        int       // group id of owner
        Size       int64     // length in bytes
        ModTime    time.Time // modified time
        Typeflag   byte      // type of header entry
        Linkname   string    // target name of link
        Uname      string    // user name of owner
        Gname      string    // group name of owner
        Devmajor   int64     // major number of character or block device
        Devminor   int64     // minor number of character or block device
        AccessTime time.Time // access time
        ChangeTime time.Time // status change time
        Xattrs     map[string]string
}
```

Thinking out loud here, beep boop

lib/fshash/metadata.go
```diff
@@ -0,0 +1,118 @@
...
+
+func ReadMetadata(path string, optional ...os.FileInfo) Metadata {
+	var fi os.FileInfo
+	var err error
+	if len(optional) > 0 {
```
kofalt: As of the tip of this branch, `ReadMetadata` is never called with `optional` params.
Yet it takes variadic args to `optional`, and further, only ever cares about the first one passed.

Unless there's something I'm missing, this should be changed to not be variadic.

lib/fshash/metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	hdr.AccessTime = def.Somewhen
+	return Metadata(*hdr)
+}
+
+func (m Metadata) Marshal(out io.Writer) {
+	// Encodes the metadata as a CBOR map -- deterministically; the output is appropriate to feed to a hash and expect consistency.
+	// We follow the rfc7049 section 3.9 description of "canonical CBOR": namely, map keys are here entered consistently, and in sorted order.
+	// Except when maps are representing a struct; then it's deterministic order, but specified by (fairly arbitrary) hardcoded choices.
+	// This doesn't implement `BinaryMarshaller` because we A: don't care and B: are invariably writing to another stream anyway.
+	// Note that if your writer ever returns an error, the codec library will panic with exactly that.  Yes, including `io.EOF`.
```
kofalt: May as well bump this docstring up above the declaration for godoc

Current output:

```
...
func ReadMetadata(path string, optional ...os.FileInfo) Metadata

func (m Metadata) Marshal(out io.Writer)

type Record struct {
    Metadata    Metadata `json:"m"`
    ContentHash []byte   `json:"h"`
}
...
```

lib/fshash/bucket.go
```diff
@@ -58,9 +15,30 @@ func (m Metadata) MarshalBinary() ([]byte, error) {
...
+	at least another two words of memory); and the Record objects are
+	serializable the same way for all implementations of Bucket (the iterators
+	may work differently depending on how data is heaped in the Bucket impl).
+*/
+type RecordIterator interface {
+	treewalk.Node
+	Record() Record
+}
+
+var InvalidFilesystem *errors.ErrorClass = errors.NewClass("InvalidFilesystem")
```
kofalt: I'd rather these end up in `def/errors.go` with their brethren.

lib/fshash/bucket.go
```diff
@@ -68,26 +46,18 @@ type Bucket interface {
...
 */
-var FileCollision *errors.ErrorClass = errors.NewClass("FileCollision")
-
-var _ Bucket = &MemoryBucket{}
-
-type MemoryBucket struct {
-	// my kingdom for a red-black tree or other sane sorted map implementation
-	lines []line
-}
+var FileCollision *errors.ErrorClass = InvalidFilesystem.NewClass("FileCollision")
```
kofalt: move --> `def/errors.go`

lib/fshash/bucket.go
```diff
@@ -68,26 +46,18 @@ type Bucket interface {
...
-type line struct {
-	metadata    Metadata
-	contentHash []byte
-}
+/*
+	MissingTree is reported when iteration over a filled bucket encounters
+	a file that has no parent nodes.  I.e., if there's a file path "./a/b",
+	and there's no entries for "./a", it's a MissingTree error.
+*/
+var MissingTree *errors.ErrorClass = InvalidFilesystem.NewClass("MissingTree")
```
kofalt: move --> `def/errors.go`

input/input.go
```diff
@@ -25,3 +29,7 @@ type Input interface {
 	*/
 	Apply(path string) <-chan error
 }
+
+var InputError *errors.ErrorClass = errors.NewClass("InputError") // grouping, do not instantiate
```
kofalt: move --> `def/errors.go`

input/input.go
```diff
@@ -25,3 +29,7 @@ type Input interface {
...
+
+var InputHashMismatchError *errors.ErrorClass = InputError.NewClass("InputHashMismatchError")
```
kofalt: move --> `def/errors.go`

lib/fshash/bucket_memory.go
```diff
@@ -0,0 +1,51 @@
...
+	"sort"
+	"strings"
+
+	"polydawn.net/repeatr/lib/treewalk"
+)
+
+var _ Bucket = &MemoryBucket{}
+
+type MemoryBucket struct {
+	// my kingdom for a red-black tree or other sane sorted map implementation
```
kofalt: picgifs-doctor-who-1512227

lib/fshash/bucket.go
```diff
@@ -68,26 +46,18 @@ type Bucket interface {
...
 
-type linesByFilepath []line
+// for sorting
+type linesByFilepath []Record
```
kofalt: This looks like a perfectly reasonable struct, but I feel it ought to be exported.

Mainly convenience, sure, but I can see plenty of implementations wanting to use.

lib/fshash/bucket_memory.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"sort"
+	"strings"
+
+	"polydawn.net/repeatr/lib/treewalk"
+)
+
+var _ Bucket = &MemoryBucket{}
+
+type MemoryBucket struct {
```
kofalt: Thoughts on moving MemoryBucket to a separate package?

Would enforce the comment I made above about implementations + imports, etc.

lib/fshash/bucket_memory.go
```diff
@@ -0,0 +1,51 @@
...
+
+type memoryBucketIterator struct {
+	lines []Record
+	this  int  // pretending a linear structure is a tree is weird.
+	that  *int // this is the last child walked.
+}
+
+func (i *memoryBucketIterator) NextChild() treewalk.Node {
+	// is the next one still a child?
+	next := *i.that + 1
```
kofalt: Okay, what. 

Is a MemoryBucket arranged in breadth-first order? Docs plz.

lib/fshash/bucket_memory.go
```diff
@@ -0,0 +1,51 @@
...
+	if next >= len(i.lines) {
+		return nil
+	}
+	if strings.HasPrefix(i.lines[next].Metadata.Name, i.lines[i.this].Metadata.Name+"/") {
```
kofalt: So, a wrench: are we assuming that filepath separator for all filesystems?
Acceptable to say "yes" for now, but should be doc'd.

lib/fshash/bucket.go
```diff
@@ -58,9 +15,30 @@ func (m Metadata) MarshalBinary() ([]byte, error) {
 */
 type Bucket interface {
 	Record(metadata Metadata, contentHash []byte)
+	Iterator() (rootRecord RecordIterator)
```
kofalt: Some one-liner docstrings here would be nice, in particular clarifying (unless I misunderstand) that Iterator walks an assembled bucket for purposes of hashing after assembly.

In fact, nothing here seems to deal with the assembly of a bucket, that's punted to implementations, yes?

lib/fshash/hash.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil
+	}
+	// Traverse
+	if err := treewalk.Walk(bucket.Iterator(), preVisit, postVisit); err != nil {
+		panic(err) // none of our code has known believable error returns.
+	}
+	// Sanity check no node left behind
+	_ = upsubs.Pop()
+	if !upsubs.Empty() || !hashers.Empty() {
+		panic("goofed")
```
kofalt: Use your words >:|

kofalt:
> details of that can be **hashed** out in the future

Stop right there criminal scum

kofalt:
Lots of comments, but overall this looks like solid work! :clap: 

Some of my comment-questions were resolved as I recursively read further, but I'm leaving them as indicators were more targeted docs would be appreciated. These packages are probably going to be the densest part of repeatr, and over-commenting is definitely preferred.

lib/fshash/metadata.go
heavenlyhash: Absolutely.  And effectively, we kind of do this -- in the marshal method.  This struct might be deserving of replacement with something more "ours" in the future, totally open to that at the first sign that keeping it becomes an irritation.

kofalt:
Incidentally, if you can avoid force-pushing that would be optimal; I'm rebasing my next-in-line nsinit work on this branch to ensure minimal divergence + take advantage of these features

lib/fshash/metadata.go
heavenlyhash: Ah, I think I had some intention to refactor other uses outside of tests to take this path too and forgot about it.  Doing that now.  (varargs are because there's no way to specify a single optional arg, womp womp polymorphism.)

lib/fshash/bucket.go
heavenlyhash: You can just refer to it as '[]Record`; that's already exported.  I wouldn't even make a type for this if that wasn't how one gets sorting behavior.

Also, nothing with type `[]Record` is ever returned, and here's why: the iterator interface is supposed to make sense for things bigger than memory so we're able to drop in a boltdb (or something) backed thing for handling absurdly large datasets where the metadata gets ponderous to handle in memory.  The iterator interface can hide that.  This detail of the memory bucket can't.

lib/fshash/metadata.go
heavenlyhash: Fixed.

lib/fshash/bucket_memory.go
heavenlyhash: Eh, could?  Yeah, `memorybucket.New()` makes sense.

lib/fshash/bucket_memory.go
heavenlyhash: Docs'ing on the `Iterator()` method.

lib/fshash/bucket.go
heavenlyhash: Actually, no. I originally put that there to indicate that this field *is* serialized, but it turns out I had to do so manually anyway in order to deal with order determinism stuff.  I'm half tempted to leave it there to make the impression and/or document that that's what these fields come out as when serialized, but it is definitely a non-functional ornament.  WDYT, stay or go?

kofalt: Without your response, I'd have expected the serialization to change when I modified those annotations. IMO, just slap a modified version of your response as a comment & leave it

lib/fshash/bucket.go
kofalt: Fair enough!

lib/fshash/metadata.go
kofalt: > varargs are because there's no way to specify a single optional arg

Hah! Fair enough; a replacement for optional & overloading, of sorts.

Given this approach, however, you should **definitely** panic if `len > 1`.

lib/fshash/bucket_memory.go
kofalt: Neither of us sound like we have strong opinions on this; I'll leave it at your discretion.

lib/fshash/bucket.go
heavenlyhash: Slapped.

lib/fshash/metadata.go
heavenlyhash: added check.

lib/fshash/bucket_memory.go
heavenlyhash: ... So I was about to say looks like `os.PathSeparator` or its alias `filepath.Separator` will DTRT... but actually, no, we want the hardcoded character here.  This has to be consistent across hosts.  If we make a lunge for portability on that level we'll want to have a translation layer well before this.

lib/fshash/bucket.go
heavenlyhash: Docs'd.

lib/fshash/hash.go
heavenlyhash: made a real-boy `ProgrammerError`.

kofalt:
Latest CI failed;

```
# polydawn.net/repeatr/lib/fshash
.gopath/src/polydawn.net/repeatr/lib/fshash/metadata.go:35: cannot call non-function "github.com/spacemonkeygo/errors".ProgrammerError (type *"github.com/spacemonkeygo/errors".ErrorClass)
```

Should be a quick fix. After that LGTM

heavenlyhash:
Doing the memory bucket move to package real quick too.

heavenlyhash:
I take that back, stashing my progress on moving memorybucket to its own package; it's not critical path and there's enough work there for its own PR.  (Making the attempt revealed a bunch of tests that are overly aware of its implementation details -- which is exactly the sort of bunkum moving it to a package *should* ferret out, but it's just the tests for those packages that are in trouble, and the mess hasn't leaked out any further than that, so as long as we fix before we add another bucket implementation, we'll be good.)

Merging as soon as ci confirms green then.

PR #12, opened 2015-03-11
Nsinit executor

kofalt:
Early days, so a few things are bundled:

* A submodule update to our minor fork of `errors` to be tolerant of mundane-typed panics.
* Dispatch refactored into packages such that an executor can use them without an import cycle.
* MVP implementation of `input/tar` and `output/tar`, using exec wrapping and blind to hashes.
* A very basic nsinit executor. Assumes the existence of a `docker0` bridge for outbound networking, does not ensure the existence of a valid DNS configuration, exec wraps `nsinit`... but works great!

Unresolved:

* The executor will create a folder for each output's location. This is great for scripts that assume the existence of an output folder (so basically, every script every written :/ ) but won't be great if an output is supposed to be a single file.
* The output tar transport is kinda wacky with its transform call, and both tar transports can be replaced later with native usage of treewalk / etc.
* The Output interface method I added currently takes the rootFS path, and is expected to join the output's Location to get something real. This is largely an artefact of our earlier internal conversation + confusion, and I could be easily convinced to bring it in line with the Input interface.


lib/flak/assist.go
```diff
@@ -39,3 +40,20 @@ func GetTempDir(dirs ...string) string {
...
+	// If IOError occurs while creating tempdir, final dir was not created to clean up.
+	dir := GetTempDir(dirs...)
+
+	try.Do(func() {
+		f(dir)
+	}).Finally(func() {
+		err := os.RemoveAll(dir)
+		if err != nil {
+			// TODO: we don't want to panic here, more like a debug log entry, "failed to remove tempdir."
+			// Can accomplish once we add logging.
```
heavenlyhash: When in doubt, let's panic anyway?  Def should be a (loud) log statement; let's not swallow until then because it dramatically increases the odds we'll forget about it.

(I've spent a lot of time debugging in golang projects only to end in discovering a `// TODO: handle error` :tada:, druther avoid it out of the gate)

executor/dispatch/dispatch.go
```diff
@@ -0,0 +1,27 @@
+package executors
```
heavenlyhash: WDYT about calling this 'executordispatch'?  Package names that have no relationship to their import names cause a huge jump the mental overhead of reading for me.

Joining the last two packages is the first compromise that comes to mind that gives the package name and import name some relationship, and also addresses the irritation of three 'dispatch' packages that would force us to alias them on every import.  Open to other ideas, that's just my first.

executor/nsinit/nsinit.go
```diff
@@ -0,0 +1,131 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	// Run inputs
+	for x, input := range job.Inputs {
+		Println("Provisioning input", x+1, input.Type, "to", input.Location)
+		path := filepath.Join(rootfs, input.Location)
+
+		// Ensure that the parent folder of this input exists
+		err := os.MkdirAll(filepath.Dir(path), 0777)
```
heavenlyhash: `0755` should be a reasonable default.

heavenlyhash:
Lots of stuff that's in placeholder territory, but this gets us rolling :+1:

kofalt:
Pushed fixes; merge when CI green

PR #15, opened 2015-03-19
Integration examples

heavenlyhash:
Some starts on integration tests, including basic asset bootstrapping (which is certainly a *huge* placeholder, but gets some things to be possible today).

### DIFF
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -9,7 +9,7 @@ import (
 
 var (
 	// os flag parsing mandates the executable name
-	baseArgs = []string{ "repeatr"}
+	baseArgs = []string{"repeatr"}
 )
 
 // These can be swapped wholesale; demonstration only.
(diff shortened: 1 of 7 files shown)


cli/cli_test.go
```diff
@@ -0,0 +1,29 @@
...
+	// os flag parsing mandates the executable name
+	baseArgs = []string{"repeatr"}
+)
+
+// These can be swapped wholesale; demonstration only.
+
+func Test(t *testing.T) {
+
+	Convey("It should not crash without args", t, func() {
+		App.Run(baseArgs)
```
heavenlyhash: We should rig this up to have the os in/out/err be parameters passed all the way through so we can shunt this output to dev null (or the capture mechanism of goconvey so it shows up nicely in reports).

heavenlyhash:
This'll need a ton more iteration but I really want some of the assets at hand to continue other work, so merging anyway.

PR #16, opened 2015-03-19
Guids

heavenlyhash:
A quick little guid generating utility.  Package docs say the rest.

(Wanted shortly for jobID generation.)

### DIFF
--- a/lib/guid/guid.go
+++ b/lib/guid/guid.go
@@ -1,0 +1,87 @@
+package guid
+
+import (
+	realrand "crypto/rand"
+	"encoding/binary"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+// base32 space; case insensitive; ascii-ordered; my least favorite characters (a bunch that can all look like vertical lines, and 'u' for looking too much like 'v') removed with prejudice.
+var pushChars = [32]byte{'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'k', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'v', 'w', 'x', 'y', 'z'}
+
+const radix = 32
+
+const randLen = 16
+
+// timexxxx-randpt1x-randpt2x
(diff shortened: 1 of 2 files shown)


lib/guid/guid.go
```diff
@@ -0,0 +1,107 @@
...
+			if lastRandChars[i] < radix {
+				break
+			}
+			// increment the next byte
+			lastRandChars[i] = 0
+		}
+	} else {
+		lastPushTimeMs = timeMs
+		for i := 0; i < randLen; i++ {
+			lastRandChars[i] = byte(rnd.Intn(radix))
```
kofalt: Am I correct in thinking this has the effect of scrambling lastRandChars? If so, why?

lib/guid/guid.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+	// put random as the third part
+	for i := 0; i < 8; i++ {
+		id[size-i-1] = pushChars[lastRandChars[i]]
+	}
+	// put random as the second part
+	for i := 8; i < 16; i++ {
+		id[size-i-2] = pushChars[lastRandChars[i]]
+	}
```
kofalt: Why the two separate array copies?

heavenlyhash: Note the gap in the constants.  There's a dash between the two of them.

Which is completely, indefensibly, wildly arbitrary, as noted in the package doc block.

lib/guid/guid.go
heavenlyhash: Mostly whynot.  Guids with no random components after the first draw seem weird to me.

lib/guid/guid.go
kofalt: Gotcha.

Personally, I find my ability to quickly snap strings breaks down after about 5-6 characters, but some separators are better than none.

heavenlyhash: Hopefully you're mostly not going to be stressed, in practice.  The first 8 characters increase monotonically, so like the first three don't move much.  We could make other breaks too -- 
this being a semirandom ID generator rather than a message-bearing serialization format, there's basically zero "migration" to worry about.

PR #17, opened 2015-03-19
Chroot executor

heavenlyhash:
First draft on a chroot-based executor.  This executor will be a lot less full-featured and promise much less containment than most -- chroots are in almost no way a valid security boundary, they don't provide network sandboxes, don't provide pid sandboxes, etc -- but they work in a lot more environments than other more modern container systems, and they'll be handy as a simple baseline.

Still a ton of WIP; this won't be very useful to merge until a lot of these todo's get struck down.

### DIFF
--- a/executor/chroot/chroot_executor.go
+++ b/executor/chroot/chroot_executor.go
@@ -11,10 +11,13 @@ import (
 	"github.com/spacemonkeygo/errors"
 	"github.com/spacemonkeygo/errors/try"
 	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/executor"
 	"polydawn.net/repeatr/input"
 	"polydawn.net/repeatr/input/dispatch"
 	"polydawn.net/repeatr/lib/guid"
 )
+
+var _ executor.Executor = &Executor{} // interface assertion
 
 type Executor struct {
 	workspacePath string // default: something like '/var/lib/repeatr/executors/chroot/'.
--- a/executor/chroot/chroot_executor.go
+++ b/executor/chroot/chroot_executor.go
@@ -1,0 +1,106 @@
+package chroot
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
+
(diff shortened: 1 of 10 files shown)


heavenlyhash:
Getting rapidly less WIPy.  We now have tests covering command execution, error codes, and that more inputs can be put into the sandbox and show up in the right places in the POV of the formula's executable.

Outputs -- both the stdout capture and the more serious CAS kind -- still need a lot of improvements, but there's enough here now to get those critical tests covered, so most of that is gonna boil over into future PRs.

executor/chroot/chroot_executor.go
```diff
@@ -0,0 +1,186 @@
...
+
+type Executor struct {
+	workspacePath string // default: something like '/var/lib/repeatr/executors/chroot/'.
+}
+
+func (x *Executor) Run(formula def.Formula) (job def.Job, outs []def.Output) {
+	try.Do(func() {
+		job, outs = x.run(formula)
+	}).Catch(input.InputError, func(e *errors.Error) {
+		// REVIEW: also directly pass input/output system errors up?  or, since we may have to gather several, put them in a group and wrap them in a "prereqs failed" executor error?
```
kofalt: I'm content to let this sit until we have more opinions about logging and production usage.

executor/chroot/chroot_executor.go
```diff
@@ -0,0 +1,186 @@
...
+	// Prepare the forumla for execution on this host
+	def.ValidateAll(&formula)
+
+	// make up a job id
+	jobID := def.JobID(guid.New())
+
+	// make a rootfs in our workspace using the jobID
+	rootfsPath := filepath.Join(x.workspacePath, string(jobID))
+	if err := os.Mkdir(rootfsPath, 0755); err != nil {
+		panic(Error.Wrap(errors.IOError.Wrap(err))) // REVIEW: WorkspaceIOError?  or a flag that indicates "wow, super hosed"?
```
kofalt: Same

kofalt:
LGTM so far.

PR #18, opened 2015-03-26
Unify jobdirs

kofalt:
* Adds a `Configure(workspacePath string)` method to executor interface
* Added some quick documentation to a few files explaining expectations
* Unify executor's usage of temp job directory logic


heavenlyhash:
LGTM

PR #19, opened 2015-03-26
Better cli test

kofalt:
* Fix CLI tests to run from the 'correct' directory, to get rid of dumb `../` prefixes
* Because of this, the basic integration test can double as a human-friendly demo
* Added some demonstration instructions!


heavenlyhash:
LGTM

PR #20, opened 2015-03-26
Check all outputs from test executions

heavenlyhash:
Goconvey's default behavior is to abort as soon as one assertion fails.  Adding the `FailureContinues` parameter to a block means any failure still fails the overall block, but will continue to do other checks in the block.  Here, this is useful to see differences in stdout/err even if, e.g., the exit code was nonzero (which is in fact when those strings are the most useful!).


heavenlyhash:
... Also, added fixes for really silly failure introduced at the last minute in the last branch about chroot executors.

PR #24, opened 2015-03-26
Travis notifications

kofalt:
Yup. IRC seems to be irritable, but no reason to not quell email spam.


heavenlyhash:
handy. :+1:

heavenlyhash:
Oh, my.  We... we seem to have caught github in a race condition.  This commit is merged, by yours truly pressing the merge button in the github web ui nonetheless, and the tip hash on this branch is indeed the same hash landed to master, and yet here I sit looking at a warning that we have conflicts to resolve.  Whee.

So, uh, deleting branch; this *is* merged.

PR #25, opened 2015-03-27
More io

heavenlyhash:
More IO, and tons of work on a smarter arrangement test fixtures.

The new test system is based on spec'ing out files in memory, easily placing them in filesystem or scanning filesystem contents to get the descriptions back, and then having tunable precision comparisons.  It's also extremely terse (jump to `testutil/filefixtures/file_fixtures.go` for the diff of glory), which is fortunate because of lot of tests so far have grown really big setups and it was starting to hinder test creation.  In the future these `Fixture` objects may  also hold other self-describing data so they can be plugged into table-driven tests.

Several refactors that became apparent on the way included (filesystem walking and the metadata structures are moved into the new `lib/fs` package, and 'lib/fshash' is now *just* the hashing components plugged into the fs walker, etc)... See the individual commit log for more; those diffs should be individually readable.

:warning: WIP :warning: -- this is one big semicontinuous landslide that will also probably still yield a pure golang tar input and hashed tar output system before it comes to rest, because all these systems just have so much overlap.  Not all prior tests are replaced yet; that can be a slow churn process.

### DIFF
--- a/lib/fs/emplacements.go
+++ b/lib/fs/emplacements.go
@@ -111,6 +111,7 @@ func PlaceDirTime(destBasePath string, hdr Metadata) {
 
 func ScanFile(basePath, path string, optional ...os.FileInfo) (hdr Metadata, file io.ReadCloser) {
 	fullPath := filepath.Join(basePath, path)
+	// most of the heavy work is in ReadMetadata; this method just adds the file content
 	hdr = ReadMetadata(fullPath, optional...)
 	hdr.Name = path
 	switch hdr.Typeflag {
--- a/lib/fs/emplacements.go
+++ b/lib/fs/emplacements.go
@@ -98,3 +98,13 @@ func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
 		ioError(err)
 	}
 }
+
+// Exposed only because you're probably doing your own trees somehow, and it's
+// necessary to cover your tracks by forcing times on dirs after all children are done.
+// Symmetric params and error handling to `PlaceFile` for your convenience.
+func PlaceDirTime(destBasePath string, hdr Metadata) {
+	destPath := filepath.Join(destBasePath, hdr.Name)
+	if err := fspatch.LUtimesNano(destPath, hdr.AccessTime, hdr.ModTime); err != nil {
+		ioError(err)
+	}
+}
--- a/lib/fs/emplacements.go
+++ b/lib/fs/emplacements.go
@@ -12,13 +12,58 @@ import (
 	"polydawn.net/repeatr/lib/fspatch"
 )
 
-func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
-	// 'destBasePath' should be an absolute path on the host.
-	// 'hdr.Name' should be the full relative path of the file.
-	// if it has an absolute prefix, that's quietly ignored, and it's treated as relative anyway.
+/*
+	Places a file on the filesystem.
+	Replicates all attributes described in the metadata.
 
+	The path is the join of `destBasePath` and `hdr.Name`.
+	`hdr.Name` should be the full relative path of the file;
+	if it has an absolute prefix, that's quietly ignored, and it's treated as relative anyway.
+	`hdr.Name` should match `filepath.Clean` output, except that it must always
+	use the unix directory separator.
+
+	No changes are allowed to occur outside of `destBasePath`.
+	Hardlinks may not point outside of the base path.
+	Symlinks may *point* at paths outside of the base path (because you
+	may be about to chroot into this, in which case absolute link paths
+	make perfect sense), and invalid symlinks are acceptable -- however
+	symlinks may *not* be traversed during any part of `hdr.Name`; this is
+	considered malformed input and will result in a BreakoutError.
+
+	`destBasePath` MUST be absolute.  Isolation checks assume this, and
+	have undefined operation if this requirement is not met.
+
+	Please note that like all filesystem operations within a lightyear of
+	symlinks, all validations are best-effort, but are only capable of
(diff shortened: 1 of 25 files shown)


lib/fs/emplacements.go
```diff
@@ -0,0 +1,126 @@
...
+	"syscall"
+
+	"github.com/spacemonkeygo/errors"
+	"polydawn.net/repeatr/lib/fspatch"
+)
+
+func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
+	// 'destBasePath' should be an absolute path on the host.
+	// 'hdr.Name' should be the full relative path of the file.
+	// if it has an absolute prefix, that's quietly ignored, and it's treated as relative anyway.
```
kofalt: Bump this a few lines up to docs?

lib/fs/walk.go
```diff
@@ -0,0 +1,113 @@
...
+	it's based on `treewalk`, which means it supports both pre- and post-order
+	traversals.
+
+	All paths begin in `.`, e.g. you'll see a series like	`{".", "./a", "./a/b"}`, etc.
+	This is identical in behavior to `filepath.Clean` having been called on each path.
+
+	Symlinks are not followed.
+
+	The traversal order of siblings is *not* guaranteed, and is *not* necessarily
+	stable.
```
kofalt: Maybe a dumb question - why these two caveats?

lib/fs/walk.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/*
+	Used in the post-order visit step so we don't continuously consume more
+	memory as we walk.  `Walk()` wraps the user-defined post-visit function
+	to do this at the end.
+*/
+func (t *FilewalkNode) forgetChildren() error {
+	t.children = nil
+	return nil
```
kofalt: Any reason this function returns a nil error?

testutil/filefixtures/types.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/*
+	Create files described by the fixtures on the real filesystem path given.
+*/
+func (ffs Fixture) Create(basePath string) {
+	for _, f := range ffs.Files {
+		fs.PlaceFile(basePath, f.Metadata, bytes.NewBuffer(f.Body))
+	}
+	// re-do time enforcement... in reverse order, so we cover our own tracks
```
kofalt: Am I understanding you correctly in that "cover our own tracks" means preventing messing up a directory?
Either way, could you clarify this comment please.

lib/fs/walk.go
heavenlyhash: Posix filesystems aren't required to give you things in sorted order.  (Most of the golang readdirs functions do the sort in-application.)

Since we tend to have to do a big-picture sort any time we care, and at other times don't care at all, it seemed easy to leave it up unsorted here.

Can add a bool flag for sorting if we hit a need for it.

testutil/filefixtures/types.go
```diff
@@ -0,0 +1,184 @@
...
+		Key   string
+		Value interface{}
+	}{
+		{"Name:%q", ff.Metadata.Name},
+		{"Type:%q", ff.Metadata.Typeflag},
+		{"Perms:%q", (map[bool]interface{}{true: "-", false: ff.Metadata.FileMode()})[opts&ComparePerms == 0]},
+		{"Mtime:%q", (map[bool]interface{}{true: "-", false: ff.Metadata.ModTime.UTC()})[opts&CompareMtime == 0]},
+		{"Atime:%q", (map[bool]interface{}{true: "-", false: ff.Metadata.AccessTime.UTC()})[opts&CompareAtime == 0]},
+		{"Uid:%d", (map[bool]interface{}{true: "-", false: ff.Metadata.Uid})[opts&CompareUid == 0]},
+		{"Gid:%d", (map[bool]interface{}{true: "-", false: ff.Metadata.Gid})[opts&CompareGid == 0]},
```
kofalt: What is this wizardry.

lib/fs/walk.go
heavenlyhash: implements `WalkFunc`.

kofalt:
LGTM; after all comments are responded to it's clear to merge.

lib/fs/walk.go
heavenlyhash: er, wait, it has no such restriction, does it.  I'll remove it.

lib/fs/walk.go
kofalt: No, that makes sense. Today I Learn *Things* About Posix.

testutil/filefixtures/types.go
heavenlyhash: If you jump through `fs.PlaceDirTime` it has a longer description of why this is sadface.  I'll buff those in the docs pass i'm pilling up now...

testutil/filefixtures/types.go
heavenlyhash: TERRRRRRRRRNARRRYS.

It was way worse before: https://github.com/polydawn/repeatr/commit/3897dce9ea2fb5a3e09f7daaf8e5ef918c1ca427

kofalt: I'm so mad right now.

testutil/filefixtures/types.go
kofalt: > necessary to cover your tracks by forcing times on dirs after all children are done.

Thought so; SGTM

heavenlyhash:
giiiiiiiiiiiiiiiiiiiiiiiiithuuuuuuuuuuuuuuuuubbbbb........

prepare for this PR with all the same content to be recreated with a new number momentarily, just 'cause reasons

PR #26, opened 2015-03-27
More io

heavenlyhash:
Literally a repost of #25 because of reasons.


heavenlyhash:
The additional input and output implementations that this is working towards have not yet condensed, but this is drifting far enough that the contributions so far should probably rejoin master... everything's in at least as working of a state as before.

PR #27, opened 2015-03-28
Refactor Job into Job + JobResult

kofalt:
Jerbs. They exist.

Sometimes, Jerbs complete. Then we need to think about our jerbs.
Sometimes, they result! In things!

*cough*

This PR significantly refactors the running Job interface to be more cognizant of asynchronous execution. A Job instance can now `Wait()` to receive a JobResult, rather than just the exit code.

A default BasicJob implementation of said interface was written, and is used by both current executors for great justice. Inherent in this upgrade are things like the ability to call `Wait()` more than once.

Along the way, the chroot executor received a staggering reduction in complexity, and now shares redundant logic with the nsinit executor (and probably all future executors that deal with a system-local file system). Finally, a convention for executor functional organization occurred.


executor/chroot/chroot_executor_test.go
```diff
@@ -31,13 +31,14 @@ func Test(t *testing.T) {
...
+			e := &Executor{
 				workspacePath: "chroot_workspace",
 			}
-			So(os.Mkdir(executor.workspacePath, 0755), ShouldBeNil)
+			So(os.Mkdir(e.workspacePath, 0755), ShouldBeNil)
 
 			Convey("We should get an InputError", func() {
-				So(func() { executor.Run(formula) }, testutil.ShouldPanicWith, input.Error)
+				result := e.Start(formula).Wait()
+				So(result.Error, testutil.ShouldBeErrorClass, tar.Error)
```
heavenlyhash: This check can stay against `input.Error` -- ShouldBeErrorClass checks the hierarchy.  Future work should swim towards a test like this not importing any specific input impl so we can move it towards tables of tests.

executor/chroot/chroot_executor.go
```diff
@@ -27,191 +26,80 @@ func (e *Executor) Configure(workspacePath string) {
...
+func (e *Executor) Run(f def.Formula, j def.Job, d string) def.JobResult {
+	var r def.JobResult
 
-		// sandbox up and invoke the real job
-		// this starts the job and immediately returns with a promise
-		job = x.invokeTask(rootfsPath, formula)
+	// TODO: wrap is bananas, wat do
+	try.Do(func() {
+		r = e.Execute(f, j, d)
+	}).Catch(errors.HierarchicalError, func(err *errors.Error) {
```
heavenlyhash: TODO can go, I think this looks solved below?

`s/errors.HierarchicalError/executor.Error/`.  We can&should limit the breadth of tree of errors this can emit... things coming out of this that aren't `executor.Error` should be considered bugs and can get bundled into `executor.UnknownError`.  Probably makes sense to whitelist `input.Error` and `output.Error` the same way.

executor/chroot/chroot_executor_test.go
kofalt: Changed

executor/chroot/chroot_executor.go
kofalt: Hmm. Verbose, but works. Both executors have been upgraded accordingly.

executor/chroot/chroot_executor_test.go
```diff
@@ -101,12 +100,9 @@ func Test(t *testing.T) {
 						Entrypoint: []string{"ls", "/data/test"},
 					}
 
-					job, _ := executor.Run(formula)
+					job := e.Start(formula)
 					So(job, ShouldNotBeNil)
-					So(job.ExitCode(), ShouldEqual, 0)
-					msg, err := ioutil.ReadAll(job.OutputReader())
-					So(err, ShouldBeNil)
-					So(string(msg), ShouldEqual, "1\n2\n3\n")
```
heavenlyhash: We need these back.  This doesn't actually test anything (except directory existence, which is different than an input having *worked*) without this check.

PR #28, opened 2015-03-31
Tar output

heavenlyhash:
Add a hashing tar output implementation.

This hashing tar output implementation is happily tiny: treewalk and fshash.bucket stuff was totally reusable, as intended... so basically all that needed to happen here was wiring the file body io streams.

Interfaces extended to support the new data returns.

Includes a new test package which uses the file fixtures, and itself refers only to the output interface -- zero references to the tar (or any other) implementation.  So these can be reused by every output implementation in the future!  All hail the table-driven specs.

### DIFF
--- a/output/tar/tar_output.go
+++ b/output/tar/tar_output.go
@@ -30,14 +30,14 @@ func New(spec def.Output) *Output {
 	}
 }
 
-func (i Output) Apply(rootPath string) <-chan error {
-	done := make(chan error)
+func (i Output) Apply(rootPath string) <-chan output.Report {
+	done := make(chan output.Report)
 	go func() {
 		defer close(done)
 
 		err := os.MkdirAll(rootPath, 0777)
 		if err != nil {
-			panic(errors.IOError.Wrap(err))
+			done <- output.Report{errors.IOError.Wrap(err).(*errors.Error), def.Output{}}
 		}
 
 		// Assumes output URI is a folder. Output transport impls should obviously be more robust
@@ -51,10 +51,9 @@ func (i Output) Apply(rootPath string) <-chan error {
 
 		err = tar.Run()
 		if err != nil {
-			done <- err
+			done <- output.Report{output.UnknownError.Wrap(err).(*errors.Error), def.Output{}}
 			return
 		}
-
 	}()
 	return done
 }
--- a/output/tar/tar_output.go
+++ b/output/tar/tar_output.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 
 	"github.com/spacemonkeygo/errors"
-
+	"github.com/spacemonkeygo/errors/try"
 	"polydawn.net/repeatr/def"
 	"polydawn.net/repeatr/output"
 )
@@ -30,31 +30,41 @@ func New(spec def.Output) *Output {
 	}
 }
 
-func (i Output) Apply(rootPath string) <-chan output.Report {
+func (o Output) Apply(rootPath string) <-chan output.Report {
 	done := make(chan output.Report)
 	go func() {
 		defer close(done)
+		try.Do(func() {
+			err := os.MkdirAll(rootPath, 0777)
+			if err != nil {
+				panic(output.TargetFilesystemUnavailableIOError(err))
+			}
 
-		err := os.MkdirAll(rootPath, 0777)
-		if err != nil {
-			done <- output.Report{errors.IOError.Wrap(err).(*errors.Error), def.Output{}}
-			return
-		}
+			// Assumes output URI is a folder. Output transport impls should obviously be more robust
+			path := filepath.Join(rootPath, o.spec.Location)
+			tar := exec.Command("tar", "-cf", o.spec.URI, "--xform", "s,"+strings.TrimLeft(rootPath, "/")+",,", path)
 
-		// Assumes output URI is a folder. Output transport impls should obviously be more robust
-		path := filepath.Join(rootPath, i.spec.Location)
-		tar := exec.Command("tar", "-cf", i.spec.URI, "--xform", "s,"+strings.TrimLeft(rootPath, "/")+",,", path)
+			//  path
+			tar.Stdin = os.Stdin
+			tar.Stdout = os.Stdout
+			tar.Stderr = os.Stderr
 
-		//  path
-		tar.Stdin = os.Stdin
-		tar.Stdout = os.Stdout
-		tar.Stderr = os.Stderr
+			// exec
+			err = tar.Run()
+			if err != nil {
+				panic(err)
+			}
 
-		err = tar.Run()
-		if err != nil {
-			done <- output.Report{output.UnknownError.Wrap(err).(*errors.Error), def.Output{}}
-			return
-		}
+			// report
+			// note: indeed, we never set the hash field.  this is *not* a compliant implementation of an output.
+			done <- output.Report{nil, o.spec}
+		}).Catch(output.Error, func(err *errors.Error) {
+			done <- output.Report{err, o.spec}
+		}).CatchAll(func(err error) {
+			// All errors we emit will be under `output.Error`'s type.
+			// Every time we hit this UnknownError path, we should consider it a bug until that error is categorized.
+			done <- output.Report{output.UnknownError.Wrap(err).(*errors.Error), o.spec}
+		}).Done()
 	}()
 	return done
 }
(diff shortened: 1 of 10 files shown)


heavenlyhash:
Added a dir output implementation because it's basically no additional work after all that foundational work.

kofalt:
I'll review later this afternoon.

Note that you've got a failing test on that last commit.

heavenlyhash:
Heh.  note to self: when creating a new feature in <2minutes, do at least watch for the CI result...

Fortunately it was just test setup PEBACs  Fixed.

This should probably sprout some more tests still, too.  Integration sanity checks against tar outputs being understood by an exec into gnu tar, for example.  Though that might come out a little graceless right now on the isolation front (until we're ready to use containment recursively in tests, which... not today).

output/tar/tar_output.go
```diff
@@ -51,10 +51,9 @@ func (i Output) Apply(rootPath string) <-chan error {
 
 		err = tar.Run()
 		if err != nil {
-			done <- err
+			done <- output.Report{output.UnknownError.Wrap(err).(*errors.Error), def.Output{}}
 			return
 		}
-
 	}()
 	return done
```
kofalt: It looks like the tar output will never actually report if there **wasn't** an error.

heavenlyhash: I propose we either add tests or usage of the report value (there currently is none since this never, well, has had a value to report) so that appears on the radar, or ignore it until it becomes an issue on its own and try to replace it before then.  WDYT?

kofalt: So, two separate thoughts.

First, I had forgotten that we close the channel and that was previously our de-facto "report". While I had thought this was totally broken, it's actually more subtle than that.

In `provision.go`, you check for `report.Err != nil` where `report` is the result of a channel read - what is the behaviour of that check when the channel was closed without being written to? I would assume either NPE, or having a "fake" default report struct.

Second, I think it's a pretty safe assumption that exec-wrapping transports will be around for awhile, and IMO they should never go away entirely. I can envision way too many cases where our nascent transports will want to be ditched (hopefully temporarily) in the name of pragmatism, and will thus need to always ensure it's possible to Accomplish Real Things on master. This isn't anything new, just our ongoing efforts to reconcile a useful execution platform with goals of repeatable computing. 

Along these lines, I think it'd be very reasonable to expect that the `output.Apply()` interface method is expected to always Report on its channel with whatever information it has. Which is this case means adding a Report in the success case of a nil Error and an Output that is a byvalue-copy of its call parameter, with the Hash field explicitly set to `""`.

kofalt: ... and add a test for that. Of course.

output/tar/tar_output.go
```diff
@@ -30,14 +30,14 @@ func New(spec def.Output) *Output {
...
-	done := make(chan error)
+func (i Output) Apply(rootPath string) <-chan output.Report {
+	done := make(chan output.Report)
 	go func() {
 		defer close(done)
 
 		err := os.MkdirAll(rootPath, 0777)
 		if err != nil {
-			panic(errors.IOError.Wrap(err))
+			done <- output.Report{errors.IOError.Wrap(err).(*errors.Error), def.Output{}}
```
kofalt: I assume you want to return here?

output/tar2/tar_output.go
```diff
@@ -31,28 +35,66 @@ func New(spec def.Output) *Output {
...
+func (o Output) Apply(basePath string) <-chan output.Report {
+	done := make(chan output.Report)
 	go func() {
 		defer close(done)
 		try.Do(func() {
+			// open output location for writing
+			// currently this impl assumes a local file uri
+			file, err := os.OpenFile(o.spec.URI, os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0755)
+			if err != nil {
+				done <- output.Report{errors.IOError.Wrap(err).(*errors.Error), def.Output{}}
```
kofalt: I assume you want to return here?

output/tests/output_tests.go
```diff
@@ -0,0 +1,86 @@
...
+
+type OutputFactory func(def.Output) output.Output
+
+func CheckScanWithoutMutation(t *testing.T, kind string, newOutput OutputFactory) {
+	testutil.Convey_IfHaveRoot("Applying the output to a filesystem shouldn't change it", t, func() {
+		for _, fixture := range filefixture.All {
+			Convey(fmt.Sprintf("- Fixture %q", fixture.Name), testutil.WithTmpdir(func() {
+				subject := newOutput((def.Output{
+					Type: kind,
+					URI:  "./output.dump", // assumes your output supports local file for throwaway :/
```
kofalt: Should we 'officially' establish a repo-local folder as a dumping ground for random stuff? 
Any allergy to using one of our tempdir functions to hide it away?

heavenlyhash: `testutil.WithTmpdir`, line 25.

The issue the comment is referring to is the presumption that whatever output implementation is being fed to this test, it has a concept of local files at whatsoever (and that a URI without a protocol is an acceptable way to specify that!).  So far this is true.  Probably won't always be.

heavenlyhash: Though I think that temp function probably hides a minor todo of its own (I copypasted it from an even older project and I'm pretty sure I didn't notice `os.Tempdir` existed at the time and so accidentally rerolled it -.-)

output/tar2/tar_output.go
heavenlyhash: Fixed (both).  Good eye.

I wish I had a clever way to make a worker pattern with an `abort(error)` pattern that would wrap the report and do the goroutine halt at the same time so a slip up like this isn't possible, but I keep coming up with ideas that seem to want generics, so... ヽ(´ー｀)ノ  gonna put that out of mind.

kofalt: You could use a further anon closure that returns a Report and send that along. Shrug.

output/tests/output_tests.go
kofalt: Fair.

kofalt:
I agree with your interface refactor 300%, it fixes an issue I knew we had to tackle with sooner or later. 

The Report issue is all that remains

output/tar2/tar_output.go
heavenlyhash: Ooh.  That's... way less bad.

output/tar/tar_output.go
kofalt: I see that a test got added for execing tar; are you comfortable with my Report thoughts above? We discussed out-of-band that an anon func for channel-returns might work well here; I'd be comfortable adapting each output implementation to 

1. Use that pattern
2. Always send a Report, regardless of success or fail

Let me know if you'd like to punt that to me, do it personally, or discuss this further.

heavenlyhash: Yep, since it's a value channel (rather than channeling a pointer), nil is not possible, and a read returning from a closed channel will use the "zero" value: http://golang.org/ref/spec#Receive_operator  So there will be no panic on those checks.

Question: *Should* the error cases also be required to return the filled-in output?  Could go either way; by comparison, if this was a multi-return method and could produce a `*def.Output`, we'd surely return nil.

Trying that extra function with a return value, but it came out pretty ugly and made me notice some other incongruities with our error handling.  We have errors being sent directly into the report channel that don't pass through the type filtering the catch blocks, and that seems wrong.  So I think I'm going to switch everything to panics inside the `try` block, and that kills both birds with one stone... except for making sure there's a success return, but that should consistently be a `done <- output.Report{nil, o.spec}` in the last line, and I can't actually figure out a way to make that less copypasta.  (...Maybe we should add a `Success`-only handler to `errors/try`? o.o  But naw, that's isomorphic to the last lines of the try block no matter what.)

heavenlyhash:
This last commit was a thorough review of error handling consistency.  All `Output` implementations now use the same `try` pattern and the same whitelisting.  I realized previous direct usage of the done channel was making it easy to slip up and send irregular errors, so that is now shunned, and in correcting it I created a new (actually meaningful!) error types.

There are also some fixmes added.  The prior work on fshash didn't return well typed errors either, which in retrospect was obviously a mistake.  I'm not sure I wanna tackle that yet.  I like the idea of an error clearly indicating whether it was the scanning part or the output committing part that had an error, but kinda wanna let that idea stew for awhile before implementing it everywhere.  For now, at least those errors all hit the "unknown" path and still get categorized, as intended.

output/tar/tar_output.go
kofalt: I'm inclined to say that when errors are a struct member, you're expected to treat the rest with scepticism + care if the error is non-nil. So I agree with your choice in later commits, I think this is fine for now.

kofalt:
Aces. LGTM

PR #29, opened 2015-04-02
Schedulers!

kofalt:
Branch name is a slight misnomer, as two schedulers landed.

* Added some docs to Executor interface methods.
* Designed a nascent Scheduler interface
* Landed a linear scheduler. Runs one job at a time.
* Landed a group scheduler. Runs N jobs at a time, defaulting to the number of virtual CPU.

The linear scheduler was originally intended as a most-basic example of how to write a scheduler. Previous comments to the effect of "why not just use group with N=1" are not wrong; I would like to keep linear around regardless to ensure we always have a dead-simple reference. And, well, >1 implementations of an interface is better than one. Open to disagreement.

At minimum, left to do:

* Executor interface proposal was agreed on, take action

I wanted to get this into PR status so we can catch anything else.

### DIFF
--- a/scheduler/group/group.go
+++ b/scheduler/group/group.go
@@ -21,11 +21,11 @@ type hold struct {
 
 type Scheduler struct {
 	groupSize int
-	executor  *executor.Executor
+	executor  executor.Executor
 	queue     chan *hold
 }
 
-func (s *Scheduler) Configure(e *executor.Executor) {
+func (s *Scheduler) Configure(e executor.Executor) {
 	s.groupSize = runtime.NumCPU()
 	s.executor = e
 	s.queue = make(chan *hold)
@@ -57,7 +57,7 @@ func (s *Scheduler) Schedule(f def.Formula) (def.JobID, <-chan def.Job) {
 func (s *Scheduler) Run() {
 	for h := range s.queue {
 
-		job := (*s.executor).Start(h.forumla, h.id)
+		job := s.executor.Start(h.forumla, h.id)
 		h.response <- job
 		job.Wait()
 	}
--- a/scheduler/group/group.go
+++ b/scheduler/group/group.go
@@ -1,6 +1,8 @@
-package linear
+package group
 
 import (
+	"runtime"
+
 	"polydawn.net/repeatr/def"
 	"polydawn.net/repeatr/executor"
 	"polydawn.net/repeatr/scheduler"
@@ -16,17 +18,21 @@ type hold struct {
 }
 
 type Scheduler struct {
-	executor *executor.Executor
-	queue    chan *hold
+	groupSize int
+	executor  *executor.Executor
+	queue     chan *hold
 }
 
 func (s *Scheduler) Configure(e *executor.Executor) {
+	s.groupSize = runtime.NumCPU()
 	s.executor = e
 	s.queue = make(chan *hold)
 }
 
 func (s *Scheduler) Start() {
-	go s.Run()
+	for w := 1; w <= s.groupSize; w++ {
+		go s.Run()
+	}
 }
 
 func (s *Scheduler) Schedule(f def.Formula) <-chan def.Job {
--- a/scheduler/group/group.go
+++ b/scheduler/group/group.go
@@ -25,10 +25,10 @@ type Scheduler struct {
 	queue     chan *hold
 }
 
-func (s *Scheduler) Configure(e executor.Executor) {
+func (s *Scheduler) Configure(e executor.Executor, queueSize int) {
 	s.groupSize = runtime.NumCPU()
 	s.executor = e
-	s.queue = make(chan *hold)
+	s.queue = make(chan *hold, queueSize)
 }
 
 func (s *Scheduler) Start() {
@@ -46,9 +46,12 @@ func (s *Scheduler) Schedule(f def.Formula) (def.JobID, <-chan def.Job) {
 		response: make(chan def.Job),
 	}
 
-	go func() {
-		s.queue <- h
-	}()
+	// Non-blocking send, will panic if scheduler queue is full
+	select {
+	case s.queue <- h:
+	default:
+		panic(scheduler.QueueFullError)
+	}
 
 	return id, h.response
 }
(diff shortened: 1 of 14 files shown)


scheduler/group/group.go
```diff
@@ -0,0 +1,61 @@
...
+func (s *Scheduler) Schedule(f def.Formula) <-chan def.Job {
+
+	h := &hold{
+		forumla:  f,
+		response: make(chan def.Job),
+	}
+
+	go func() {
+		s.queue <- h
+	}()
```
heavenlyhash: Let's make this a queue.  It's currently not: this is a rube-goldberg use of goroutines that creates an implicit and unordered queue based on blocking otherwise pointless goroutines.

Making a buffered channel will actually perform the role of a queue.

We should also define behavior for when the queue is full.  This is an inevitable endgame for scheduler designs and does not get simpler later.  http://mechanical-sympathy.blogspot.co.uk/2012/05/apply-back-pressure-when-overloaded.html provides a good background on why; the "Rejected tasks" heading in https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html provides some very useful concrete examples of how other battle-hardened systems can deal with this.  For now a function for strategy and a default implementation that panics is probably fine, and is better than ignoring it.

scheduler/group/group.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return h.response
+}
+
+// Run jobs one at a time
+func (s *Scheduler) Run() {
+	for h := range s.queue {
+		id := def.JobID(guid.New())
+		job := (*s.executor).Start(h.forumla, id)
```
heavenlyhash: Unnecessary `*`s...?

kofalt:
Rebased onto master and reviews complete; this is ready for review.

The last two commits in particular are worth individual attention as they respectively:

1. Require that executors be handed a guid (proposal previously discussed out-of-band)
1. Take that change to its IMO logical conclusion - schedulers now immediately return said guid

There's more than one way to approach that second point; my reasoning is the following:

* Job references currently hold an `io.Reader` and may someday hold more, much of which should definitely be punted to an executor's implementation / discretion.
* Thus, we don't want to instantiate a Job for the executor
* Schedulers may have to wait arbitrarily long before executing a formula
* Having no way to unique identify a queued forumla until it actually *starts* is sloppy

Having the scheduler generate the guid and hand it back was the quickest way to deal with this.

Possible alternatives:

1. The executors interface gains a method that asks executors to instantiate a Job that the scheduler can call immediately. Set up buffers or whatever; then remove the guid and add a Job to `executor.Start()`. Explicitly document expectation that reading from the OutputReader can block for arbitrarily long time.
1. The scheduler creates the Job and hands it off; Job interface gains a `WaitForStart()` or something that you are expected to call before trying to read from the uninitialized reader.
1. ???

Notice how these kinds feel like poor excuses for us lacking a state machine? IDK. Thoughts?

scheduler/scheduler.go
```diff
@@ -0,0 +1,50 @@
...
+	These Jobs might be scheduled locally or over an enormous cluster of remote machines.
+
+	Because these imply significant complexity, it would be naive to assume that a scheduler interface could be one-size-fits-all.
+	This interface comprises an optimistic starting point that schedulers can follow, but may surpass.
+
+	Schedulers are presumed to know environmental context that a Formula provider may not.
+	Any problems with a transport running on a remote host, for example, is the scheduler's problem.
+
+	REVIEW: our interfaces do not support returning a guid from scheduling, as executors create their guid when starting.
+	It would be cleaner to be able to refer to a scheduled Job without waiting arbitrarily long for the scheduler to start the job.
```
heavenlyhash: I'm all in favor of this: setting up the guid so we can talk about a job before it's been launched is totally correct.  Looks like you already made this change to Executors... should it push out through schedulers too?  I'm not really sure, since we haven't doc'd what exactly the scheduler's responsibility level is -- is it just supposed to handle pushing tasks out to a farm, or is it actually the originator of work requests?  Probably the latter, thinking out loud, which would indicate yeah, let's push the guid assignment one level higher yet.

heavenlyhash:
> poor excuses for us lacking a state machine

Yep, basically.

I think pushing guid assignments out is the right incremental step right now, and yes, we should plan on defining a state machine for what level of "planned" versus running a job is in.  Maybe soon, given this experience.

scheduler/group/group.go
kofalt: Agreed on rube-goldberg status; this is punting pretty hard and will fall apart when things go #webscale, where #webscale is whenever the golang scheduler starts calling my bluff on goroutine madness. From anecdotal personal experience I know that is at least in the tens of thousands, and presumably capable of a few more orders of magnitude. 

Long-term, I'd like both ordering and introspection capability on the queue, only one of which is provided by buffered channels. A steady-state for scheduling interface design is likely to be a long process, so your suggestion sounds fine in the interim.

I'd be comfortable with this PR setting the channels to be buffered, add a Configure param for queue size, and perhaps panicing with a `QueueFullError` if the job cannot be enqueued.

scheduler/group/group.go
kofalt: Looks like a remainder from type system shenanigans. Fixed.

scheduler/scheduler.go
kofalt: Agreed and completed in  91100a4

scheduler/group/group.go
heavenlyhash: +1, that'll do.  I don't think there's any reason not to set the buffer size in the thousands, either (afaik there's no serious pre-allocations, though maybe double check me).

(Tangentially now: "tens of thousands" can be situational.  Tens of thousands of goroutines floating "somewhere"?  Fairly fine.  Tens of thousands of them blocking on a single unbuffered queue, forcing all of them to be considered during a single wakeup?  Might be hitting paths that force the scheduler to do a lot of rethinks and get `O(painful)` at much lower numbers.)

I think the spec says you can actually `len()` on a chan.  Not saying it's going to be "thread-safe" (making choices on it instead of an attempt-and-handle-errors would be a TOCTOU), but it does offer rough inspection.

kofalt: Nice, the `len()` thing may be enough for initial efforts. I'll make the changes we've agreed on, and I think that'll be it for this PR.

kofalt:
Updated; merge if tranquil.

heavenlyhash:
Tranquility affirmed.

PR #30, opened 2015-04-07
Stream muxer

heavenlyhash:
Add a stream logging and mux/demuxing system.

After creating a mux, one can get labeled writers to it, and readers that return the conjunction of any choice of labels.  Every reader starts off again from the beginning.  This implementation is backed by files, so it's safe for arbitrarily large amounts of data.  Readers can safely run concurrently with writers; the readers will return as much as they can, and when out of data will continue to block until the writers emit close signals (even if the reader catches up with the current end of file while any relevant writer is still ongoing).

The intended/example use case for this is collecting stdout and stderr from a process, saving both of them, and being able reply either alone or both together, always in the same order they originally appeared.

There's a couple todo's scattered around; we can decide how important those are to fix before merging.  There's numerous sharp edges to this code (for one example, asking for streams labels that never existed means you'll block forever waiting for them to be "closed"), but if used as directed should be functional.

### DIFF
--- a/lib/streamer/cbor_streamer.go
+++ b/lib/streamer/cbor_streamer.go
@@ -1,0 +1,136 @@
+package streamer
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"os"
+	"sync"
+
+	"github.com/spacemonkeygo/errors"
+	"github.com/ugorji/go/codec"
+)
+
+var _ Mux = &CborMux{}
+
+type CborMux struct {
+	// the fd tip handles append; reads all use ReadAt and track their offsets individually
+	file  *os.File
+	codec *codec.Encoder
+	wmu   sync.Mutex
+}
+
+func CborFileMux(filePath string) Mux {
+	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_RDWR|os.O_EXCL, 0755)
+	if err != nil {
+		panic(errors.IOError.Wrap(err))
+	}
+	file.Write([]byte{codec.CborStreamArray})
+	return &CborMux{
+		file:  file,
+		codec: codec.NewEncoder(&debugWriter{file}, new(codec.CborHandle)),
+	}
+	// consider using runtime.SetFinalizer to close?  currently unhandled.
+}
+
+func (m *CborMux) write(label int, msg []byte) {
+	m.wmu.Lock()
+	defer m.wmu.Unlock()
+	const magic_RAW = 0
+	_, enc := codec.GenHelperEncoder(m.codec)
+	enc.EncodeMapStart(1)
+	enc.EncodeInt(int64(label))
+	enc.EncodeStringBytes(magic_RAW, msg)
+	// REVIEW: we might need an extra flag for "close"
+	// could sneak it in as a different type here?  might be rude on parsers.
+}
+
+func (m *CborMux) Close() {
+	m.wmu.Lock()
+	defer m.wmu.Unlock()
+	m.file.Write([]byte{0xff}) // should be `codec.CborStreamBreak`, plz update upstream for vis
+	// don't *actually* close, because readers can still be active on the same fd.
+	// TODO further writes should be forced into an error state
+}
+
+func (m *CborMux) Appender(label int) io.WriteCloser {
+	return &cborMuxAppender{m, label}
+}
+
+type cborMuxAppender struct {
+	m     *CborMux
+	label int
+}
+
+func (a *cborMuxAppender) Write(msg []byte) (int, error) {
+	a.m.write(a.label, msg)
+	return len(msg), nil
+}
+
+func (a *cborMuxAppender) Close() error {
+	// TODO write a special flagged event?
+	return nil
+}
+
+func (m *CborMux) Reader(labels ...int) io.Reader {
+	// asking for a reader for a label that was never used will never
+	// hit a close flag, so... don't do that?
+	r := io.NewSectionReader(m.file, 1, math.MaxInt64/2)
+	r2 := &debugReader{r}
+	// TODO offset of one because that's the array open!
+	// do something much more sane than skip it, please
+	return &cborMuxReader{
+		labels: labels,
+		codec:  codec.NewDecoder(r2, new(codec.CborHandle)),
+	}
+}
+
+type debugReader struct {
+	r io.Reader
+}
+
+func (r *debugReader) Read(b []byte) (int, error) {
+	i, e := r.r.Read(b)
+	fmt.Printf(":: dbg read: %#v    %#v    %#v\n", i, b, e)
+	return i, e
+}
+
+type debugWriter struct {
+	w io.Writer
+}
+
+func (w *debugWriter) Write(b []byte) (int, error) {
+	i, e := w.w.Write(b)
+	fmt.Printf(":: dbg write: %#v    %#v    %#v\n", i, b, e)
+	return i, e
+}
+
+type cborMuxReader struct {
+	labels []int // remove them as we hit their close
+	codec  *codec.Decoder
+}
+
+func (r *cborMuxReader) Read(msg []byte) (int, error) {
+	var row map[int]interface{}
+	err := r.codec.Decode(&row)
+	if err == io.EOF {
+		return 0, err
+	}
+	fmt.Printf("::: read row: %#v\n", row)
+	// read the map bits
+	// read the key
+	// switch on the value type
+	//	switch {
+	//	case dec.IsContainerType(valueTypeInt):
+	//	case dec.IsContainerType(codec.valueTypeBytes)
+	//	}
+	for k, v := range row {
+		// this could be wildly more efficient with more handcoded parsing
+		for _, l := range r.labels {
+			if k == l {
+				return copy(msg, v.([]byte)), nil
+			}
+		}
+	}
+	return 0, nil
+}
--- a/lib/streamer/cbor_streamer.go
+++ b/lib/streamer/cbor_streamer.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/spacemonkeygo/errors"
 	"github.com/ugorji/go/codec"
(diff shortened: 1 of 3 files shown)


lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+	Msg   []byte `json:"m,omitempty"`
+	Sig   int    `json:"x,omitempty"` // 1->closed
+}
+
+func CborFileMux(filePath string) Mux {
+	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_RDWR|os.O_EXCL, 0755)
+	if err != nil {
+		panic(errors.IOError.Wrap(err))
+	}
+	file.Write([]byte{codec.CborStreamArray})
```
kofalt: This is the byte header to indicate the beginning of a cbor stream?

lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+	//	m.codec.MustEncode(cborMuxRow{
+	//		Label: label,
+	//		Msg:   msg,
+	//	})
+	_, enc := codec.GenHelperEncoder(m.codec)
+	enc.EncodeMapStart(2)
+	enc.EncodeString(magic_UTF8, "l")
+	enc.EncodeInt(int64(label))
+	enc.EncodeString(magic_UTF8, "m")
+	enc.EncodeStringBytes(magic_RAW, msg)
```
kofalt: I was reminded about [GenHelperEncoder](https://godoc.org/github.com/ugorji/go/codec#GenHelperEncoder) when looking it up again:

> DO NOT USE IT DIRECTLY. IT WILL CHANGE CONTINOUSLY WITHOUT NOTICE. 

I assume this is in use because you needed strict key ordering, as with the filesystem stream? 
Wasn't that ticket resolved (and either way, link me please)? Or is this something else?

lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+	m.wmu.Lock()
+	defer m.wmu.Unlock()
+	const magic_RAW = 0
+	const magic_UTF8 = 1
+	//	m.codec.MustEncode(cborMuxRow{
+	//		Label: label,
+	//		Msg:   msg,
+	//	})
+	_, enc := codec.GenHelperEncoder(m.codec)
+	enc.EncodeMapStart(2)
```
kofalt: Just checking as the I dead-ended when looking up the [interface](https://github.com/ugorji/go/blob/11c92bbf14fb06664d2a8ad2156de0ee58c51758/codec/encode.go#L53):

This is indicating merely a map of 2 keys?

lib/streamer/cbor_streamer.go
heavenlyhash: Yep. An indefinite length array, to be precise.

Currently it's never closed.  One of the rough edges that should maybe be fixed.  Design blocker for doing so sanely: would have to change the mux interface to have a known-ahead-of-time set of labels.  Right now, all the readers just leave when they hit the close signal for their labels, so nobody's ever around to see the stream (not) end correctly.

lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+	enc.EncodeInt(int64(1))
+	return nil
+}
+
+func (m *CborMux) Reader(labels ...int) io.Reader {
+	// asking for a reader for a label that was never used will never
+	// hit a close flag, so... don't do that?
+	r := io.NewSectionReader(m.file, 1, math.MaxInt64/2)
+	// TODO offset of one because that's the array open!
+	// do something much more sane than skip it, please
```
heavenlyhash: todo: I think I recall an optional encoding type hint byte as well, which did for cbor what checking for a leading "{" does for json.  It would make sense to put that in the front of this stream too.  Can't remember where I saw it or what it's called atm

lib/streamer/cbor_streamer.go
heavenlyhash: I believe it was "canonical encodings", issue 56.  Upstream added support for deterministic encodings using sorted order (which is cool), but didn't expose direct choice of ordering.  We could go about our business without using this interface, but I dunno, despite the scare-caps, it doesn't really seem too worrisome (we've got pegged deps et all).

lib/streamer/cbor_streamer.go
heavenlyhash: Spot on.

lib/streamer/cbor_streamer.go
kofalt: Confirmed ([link](https://github.com/ugorji/go/issues/56))

Do we care about the order as long as it's consistent? Are you continuing to use `GenHelperEncoder` for this & the filesystem stream to avoid the developer-mentioned overhead of encoding out-of-band + compare?

I just want to make sure there's a good reason to keep using this interface.

lib/streamer/cbor_streamer.go
kofalt: It looks like you've designed labels to independently drop in & out.
Given our use case, wouldn't we expect all labels to close around the same time?
Do you envision a case where we'll need this flexibility?

One implementation option: `CborMux.Appender(label int)` remembers the label added in a set that's kept on the CborMux and protected by the same mutex. Then `cborMuxAppender.Close()` would additionally call a new method `CborMux.CloseAppender(label int)` subtracting the label from the set.

The `CborMux` would emit a stream close header when `CloseAppender` is called and results in an empty set of open labels. Any attempts to write after that occurs would be an error.

lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+func (a *cborMuxAppender) Close() error {
+	a.m.wmu.Lock()
+	defer a.m.wmu.Unlock()
+	const magic_UTF8 = 1
+	_, enc := codec.GenHelperEncoder(a.m.codec)
+	enc.EncodeMapStart(2)
+	enc.EncodeString(magic_UTF8, "l")
+	enc.EncodeInt(int64(a.label))
+	enc.EncodeString(magic_UTF8, "x")
+	enc.EncodeInt(int64(1))
```
kofalt: At time of writing, is it true that the first instance of a label map that defines `x` to be the close map and the last map ever written under that label?

lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+	buf    []byte // any remaining bytes from the last incomplete read
+}
+
+func (r *cborMuxReader) Read(msg []byte) (n int, err error) {
+	n, err = r.read(msg)
+	for n == 0 && err == nil {
+		// we're effectively required to block here, because otherwise the reader may spin;
+		// this is not a clueful wait; but it does prevent pegging a core.
+		// quite dumb in this case is also quite fool-proof.
+		time.Sleep(1 * time.Millisecond)
```
kofalt: Could I get your brief thoughts about upgrades to this in writing?
We discussed this in person but I think having a driveby of future ideas recorded would be wise.

I'm totally fine with landing bit, if that wasn't clear. Motivation is one of remembering

lib/streamer/cbor_streamer.go
heavenlyhash: There's no requirement that processes close their starting file descriptors all at once.  So inevitably, they don't, and Something is going to rely on that.  I remember finding this out the hard way while working on siphon at some point, though I don't think I was decent enough at the time to write tests or examples.

I guess we don't really *need* the array close flag in order to be valid cbor, anyway?  Yeah, check out around the `dec.MustDecode(&reheated)` line in the tests... apparently that's parsing validly.
Something like that would work, but adding new labels to the open set while other goroutines are already allowed to be closing them is going to make a big picture race, so either the mux interface should start looking like a factory to make that clear, or we're gonna get a rather sharp "safe" usage path where the caller must make sure to open all appenders before possibly closing any of them.

lib/streamer/cbor_streamer.go
```diff
@@ -0,0 +1,199 @@
...
+	so that higher level consumers of the Reader interface don't get stuck
+	spin-looping.
+
+	Specifically, these situations cause empty reads:
+	  - hitting EOF on the backing file, but still having labelled streams
+	    that haven't been closed (i.e. we expect the file to still be growing)
+	  - absorbing a message that isn't selected by this reader's filters
+	  - absorbing a message that's a signal and has no body
+*/
+func (r *cborMuxReader) read(msg []byte) (int, error) {
```
kofalt: Does this method buffer the raw bytewise reads and only omit complete 'rows'?

lib/streamer/cbor_streamer.go
heavenlyhash: Two "smarter" routes exist:

- Have an atomic flag and wait in-process
- Use fnotify

The former breaks down if we ever want this to work across processes (seems like it might happen); the latter breaks down kind of whenever (afaict ever fnotify library on the planet falls back to polling like this again, because there's fd limits and fnotify and all sorts of edge cases).

So this is "stupid", but it's the one thing that universally Works.

lib/streamer/cbor_streamer.go
kofalt: Hmm. I imagine the hard way is the only way to learn that tidbit :)

General agreement on race. I was assuming that we'd be setting up all our labels before actually handling the mux off to a process (aka a Gosh template), at which point that becomes a compelling case for demanding to know the full set of labels ahead of time, anyway.

Tackling this from the other direction, then, will it be in poor taste to mandate a known-ahead-of-time set of labels? Do you plan to head in that direction?

kofalt: I'm going to claim that this [spec section](https://tools.ietf.org/html/rfc7049#section-2.2.1) appears to give us clearance to not have a terminus. I would feel better if we had one though - it feels... more competent? ... to be able to definitely know when we're done. Will probably also be useful if these streams end up getting sent over something with a more complex EOF reality (network).

heavenlyhash: ...Yeah, that about sums up my feelings after additional mulling too.  Nagging desire to not leave that dangling.

heavenlyhash: Will putter with interface, not sure what will turn out to be min-ulgy.  (I think it's stuck between an asymmetrical interface (one method is setup, other method is use, do not mix the order), or a factory method, neither of which is really attractive...)

kofalt: I would tentatively vote in the direction of asymmetric, as we have that going for Schedulers already, but I'll go along with whichever you feel works out best.

kofalt:
Based on out-of-band conversation, I'm going to merge this; I propose that you continue adding commits on this branch and we merge again when outstanding points are satisfied.

lib/streamer/cbor_streamer.go
heavenlyhash: Yeah, if do the reflection for the struct, then allocate storage space for the keys, then invoke a sorting operation, then starte serializing, every time something does an `fwrite` to stdout, it sounds like quite a bit of jig, neh?

We *could* use the canonical feature; but this does seem like an "in a tight loop" sort of scenario, so I'm a fan of the direct route.  We're also free to change it later if it gets unmaintainable.

kofalt: Gotta Go Fast.

PR #31, opened 2015-04-08
Exec upgrades

heavenlyhash:
Start upgrading use of exec to use [Gosh](https://github.com/polydawn/gosh/).

Using gosh means their error handling is now consistent in all the ways gosh addresses (exit codes are accounted for, signals are accounted for, ptrace is accounted for, etc; we can remove that copyforked stuff from repeatr).   In the cases upgraded here, we also now make operations silently (not mucking up the host terminal), and if something goes wrong then (and only then) attach output to the error we're about to raise -- this fixes a few places where goconvey was reporting fails because of misdirected output.

### DIFF
--- a/input/errors.go
+++ b/input/errors.go
@@ -1,0 +1,30 @@
+package input
+
+import (
+	"github.com/spacemonkeygo/errors"
+)
+
+var Error *errors.ErrorClass = errors.NewClass("InputError") // grouping, do not instantiate
+
+/*
+	Indicates that the input failed to obtain data that correctly
+	matches the hash specifying the input.  This may mean there have
+	been data integrity issues in the storage or transport systems involved.
+*/
+var InputHashMismatchError *errors.ErrorClass = Error.NewClass("InputHashMismatchError")
+
+/*
+	Indicates that the target filesystem (the one given to `Apply`) had some error.
+*/
+var TargetFilesystemUnavailableError *errors.ErrorClass = Error.NewClass("TargetFilesystemUnavailableError")
+
+// Convenience method for wrapping io errors.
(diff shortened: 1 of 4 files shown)


kofalt:
Does this need to use the Gosh PRs that just landed / are landing?

heavenlyhash:
Nope!  There's more coming that does, but I put that into a different branch so these are ready to go in without delay.

kofalt:
LGTM

input/errors.go
```diff
@@ -0,0 +1,30 @@
...
+	Indicates that the input failed to obtain data that correctly
+	matches the hash specifying the input.  This may mean there have
+	been data integrity issues in the storage or transport systems involved.
+*/
+var InputHashMismatchError *errors.ErrorClass = Error.NewClass("InputHashMismatchError")
+
+/*
+	Indicates that the target filesystem (the one given to `Apply`) had some error.
+*/
+var TargetFilesystemUnavailableError *errors.ErrorClass = Error.NewClass("TargetFilesystemUnavailableError")
```
kofalt: We have an identically-named be string-indistinguishable error in the output package.

I merged this before realizing I hadn't left this comment, and I don't care strongly enough to hold this up.
Take some or no action as desired.

heavenlyhash: Ick, good eye.  I'm still kind of dead reckoning / flying through fog on errors big-picture-wise, and wondering if maybe expressing which relevant filesystem should just be a [DataKey](https://godoc.org/github.com/spacemonkeygo/errors#DataKey) thing?  So that's something to ponder for future error reviews&refactors... but nonunique names is offensive even in the interim.  Fixing in #32.

PR #33, opened 2015-04-09
Exec upgrades

heavenlyhash:
Upgrade chroot executor to use [https://github.com/polydawn/gosh](gosh).

Both lets us remove a bunch of duplicated exit detection code, and gets some other fixes in from gosh for things like better command-not-found edge case detection than repeatr previously had.

### DIFF
--- a/executor/chroot/chroot_executor.go
+++ b/executor/chroot/chroot_executor.go
@@ -6,9 +6,9 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"github.com/polydawn/gosh"
 	"github.com/spacemonkeygo/errors"
 	"github.com/spacemonkeygo/errors/try"
-
 	"polydawn.net/repeatr/def"
 	"polydawn.net/repeatr/executor"
 	"polydawn.net/repeatr/executor/basicjob"
@@ -69,10 +69,8 @@ func (e *Executor) Run(f def.Formula, j def.Job, d string) def.JobResult {
 func (e *Executor) Execute(f def.Formula, j def.Job, d string) def.JobResult {
 
 	result := def.JobResult{
-		ID:       j.Id(),
-		Error:    nil,
-		ExitCode: 0, //TODO: gosh
-		Outputs:  []def.Output{},
+		ID:      j.Id(),
+		Outputs: []def.Output{},
 	}
 
 	// Prepare filesystem
@@ -91,15 +89,27 @@ func (e *Executor) Execute(f def.Formula, j def.Job, d string) def.JobResult {
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
-	if err := cmd.Start(); err != nil {
-		if err2, ok := err.(*exec.Error); ok && err2.Err == exec.ErrNotFound {
+	// launch execution.
+	// transform gosh's typed errors to repeatr's hierarchical errors.
+	var proc gosh.Proc
+	try.Do(func() {
+		proc = gosh.ExecProcCmd(cmd)
+	}).CatchAll(func(err error) {
+		switch err.(type) {
+		case gosh.NoSuchCommandError:
 			panic(executor.NoSuchCommandError.Wrap(err))
+		case gosh.NoArgumentsErr:
+			panic(executor.NoSuchCommandError.Wrap(err))
+		case gosh.ProcMonitorError:
+			panic(executor.TaskExecError.Wrap(err))
+		default:
+			panic(executor.UnknownError.Wrap(err))
 		}
-		panic(executor.TaskExecError.Wrap(err))
-	}
+	}).Done()
 
 	// Wait for the job to complete
-	result.ExitCode = flak.WaitAndHandleExit(cmd)
+	// REVIEW: consider exposing `gosh.Proc`'s interface as part of repeatr's job tracking api?
+	result.ExitCode = proc.GetExitCode()
 
 	// Save outputs
 	result.Outputs = flak.PreserveOutputs(f.Outputs, rootfs)
--- a/lib/flak/assist.go
+++ b/lib/flak/assist.go
@@ -3,9 +3,7 @@ package flak
 import (
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
-	"syscall"
 
 	"github.com/spacemonkeygo/errors"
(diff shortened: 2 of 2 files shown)


executor/chroot/chroot_executor.go
```diff
@@ -91,15 +89,29 @@ func (e *Executor) Execute(f def.Formula, j def.Job, d string) def.JobResult {
...
+			panic(executor.NoSuchCommandError.Wrap(err))
+		case gosh.ProcMonitorError:
+			panic(executor.TaskExecError.Wrap(err))
+		default:
+			panic(executor.UnknownError.Wrap(err))
 		}
-		panic(executor.TaskExecError.Wrap(err))
-	}
+		// FIXME all *this* also should actually be done in the barrier before returning a job, methinks?
+		// actually, not clear.  we haven't really spec'd that.  REVIEW.
```
kofalt: Yeah, Job instantiation problems intensify. I want to work on that separately.

IMO this should stay for now; we'll work on Jobs separately.

executor/chroot/chroot_executor.go
```diff
@@ -91,15 +89,29 @@ func (e *Executor) Execute(f def.Formula, j def.Job, d string) def.JobResult {
...
+	}).Done()
 
 	// Wait for the job to complete
-	result.ExitCode = flak.WaitAndHandleExit(cmd)
+	// REVIEW: consider exposing `gosh.Proc`'s interface as part of repeatr's job tracking api?
+	result.ExitCode = proc.GetExitCode()
```
kofalt: I think we might end up more in the business of implementing something that looks very like the [Proc interface](https://godoc.org/github.com/polydawn/gosh#Proc), honestly. I want a progress-report concept of state that has more steps than just [these states](https://godoc.org/github.com/polydawn/gosh#State) for example.

Oh and by the way we have one struct to report on this but really the Executor ought to stay in control of that. More fuel for my thought of the Executor gaining a method to build itself a Job way before the Scheduler wants to run a Formula, and the Job gets better at saying what state it's in.

I propose that, ignoring gosh, we figure out how a running Job should really look, and then decide what parts (if any) of Gosh bleed through. I might have a dedicated branch to poke at that after we "clear the board" a bit.

kofalt:
This LGTM so far and I want to tackle your review points (that I think we've both know have been coming) separately. Let me know if disagree?

Do you plan to upgrade nsinit as well?

executor/chroot/chroot_executor.go
heavenlyhash: A+

If anything I was thinking letting Proc/State from gosh be used to describe execution state once a job has been queued into a machine... but yes, there's definitely a bunch of other concepts there we need to draw out, and definitely a larger statemachine.

executor/chroot/chroot_executor.go
heavenlyhash: I'm coming around to thinking it's totally fine and I'll remove the comment.  It doesn't make any sense at all to have executors be generally presumed to have *launched* a computation synchronously or to have sanity checked command existence.

executor/chroot/chroot_executor.go
kofalt: Agreed.

My current plan is to land `use-mux-in-executors` after a rebase onto this PR, and then a new branch specifically to mess with Jobs. `use-mux-in-executors` will perpetuate or exacerbate the problem because I really want to mess with Jobs separately from mux upgrades, etc.

heavenlyhash:
I looked at changing the nsinit executor, but I backed off because I'm not actually sure what to do with it.  I'm not sure nsinit proxies exit codes (and if so, it's automatically not lossless because nsinit errors gotta go somewhere), and similarly command-not-found for the job definition is a totally different thing than command-not-found for nsinit when setting up.

I don't know if we're handling either of those situations now either, but I'm a lil wary of committing new code that's still muddled on those, and also just plain of changing a thing that hasn't got tests.  And at any rate, it's not going to follow the outline of the chroot executor on this part.

executor/chroot/chroot_executor.go
heavenlyhash: Sage ordering.  ...Tangent, now, but do you know if you're going to end up touching any of the flak functions executors call right now?  I have reshuffling them to shush IO getting near the top of my todo list.

executor/chroot/chroot_executor.go
heavenlyhash: ... I was going to add more docs to the `Executor` interface to cover this, but actually...

> An invalid formula (as far as can be determined without hitting the filesystem) will panic before returning.

Yeah, okay.  That actually covers this pretty well.

kofalt:
In that case, we might want to bring back `flak.WaitAndHandleExit` and use it [here](https://github.com/polydawn/repeatr/blob/c2968f11e27711072c8ea501925a744f63a03c9c/executor/nsinit/nsinit.go#L119)?

Up to you; modify or merge as desired.

heavenlyhash:
My feeling on that is that the area is pants and will remain pants until it gets a whole nsinit-centric makeover.  Getting those errors more correctly still won't tell a real story about the user's process.  I'm putting "retrofit chroot tests to be executor agnostic / shoop the whoop on nsinit" on the whiteboard checklist.

heavenlyhash:
:soon:, in other words.  :soon:.

kofalt:
I think Travis is having a stroke

https://travis-ci.org/polydawn/repeatr/builds/57842586
Original content; am triggering a rerun for maximum hopes & dreams

```
$ cd polydawn/repeatr

$ git fetch origin +refs/pull/33/merge:

fatal: Couldn't find remote ref refs/pull/33/merge

The command "eval git fetch origin +refs/pull/33/merge:" failed. Retrying, 2 of 3.

fatal: Couldn't find remote ref refs/pull/33/merge

The command "eval git fetch origin +refs/pull/33/merge:" failed. Retrying, 3 of 3.

fatal: Couldn't find remote ref refs/pull/33/merge

The command "eval git fetch origin +refs/pull/33/merge:" failed 3 times.

The command "git fetch origin +refs/pull/33/merge:" failed and exited with 128 during .
```

heavenlyhash:
I did a rebase with nothing but the comment removed right before hitting merge and deleting the branch.  I think this is Travis not understanding branches going away being different than execution failed.

PR #34, opened 2015-04-10
Use mux in executors

kofalt:
* Hooks up the new stream mux to executors.
* Delays handing off the Job until the muxes are ready.
 * Does nothing to improve the Job situation; that's a different topic
* Now that we have nicer error handling, squelch redundant reporting
* Stream job output from mux to CLI in real time

### DIFF
--- a/executor/chroot/chroot_executor.go
+++ b/executor/chroot/chroot_executor.go
@@ -32,27 +32,42 @@ func (e *Executor) Start(f def.Formula, id def.JobID) def.Job {
 
 	// Prepare the forumla for execution on this host
 	def.ValidateAll(&f)
+
 	job := basicjob.New(id)
+	jobReady := make(chan struct{})
 
 	go func() {
 		// Run the formula in a temporary directory
 		flak.WithDir(func(dir string) {
-			job.Result = e.Run(f, job, dir)
+
+			// spool our output to a muxed stream
+			var strm streamer.Mux
+			strm = streamer.CborFileMux(filepath.Join(dir, "log"))
+
+			outS := strm.Appender(1)
+			errS := strm.Appender(2)
+			job.Reader = strm.Reader(1, 2)
+
+			// Job is ready to stream process output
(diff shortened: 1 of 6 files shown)


executor/chroot/chroot_executor.go
```diff
@@ -66,7 +82,7 @@ func (e *Executor) Run(f def.Formula, j def.Job, d string) def.JobResult {
 }
 
 // Execute a formula in a specified directory. MAY PANIC.
-func (e *Executor) Execute(f def.Formula, j def.Job, d string) def.JobResult {
+func (e *Executor) Execute(f def.Formula, j def.Job, d string, outS, errS io.WriteCloser) def.JobResult {
```
heavenlyhash: This method sig is starting to get a liiiittle scary.  Would it make sense to swap out `j`, `outS`, and `errS` for handing in a channel that's expected to yield the ready job object?

heavenlyhash: Probably depends on if the future involves any more stuff from the `Start` method's context going into the job object construction, so I guess this is season to taste.

heavenlyhash:
LGTM.  Super excited about getting my terminal clearer again during tests.

So I think... with this PR, we could now hook up the chroot executor tests to nsinit, too?  At least the output ones.  Which would be *awesome*.

executor/chroot/chroot_executor.go
kofalt: Basically, I think both Job and our currently-used BasicJob implementation should get cooler. I postponed that larger change in favour of mucking the by-convention function sigs; goal is that they'll shrink again when that's all figured out.

kofalt:
Some of them, though a few deal with exit codes & command not found, which as previously discussed might be a little ¯\_(ツ)_/¯ right now with nsinit. I'd be receptive to templating those a bit and cross-applying them.

I'll merge this as end-to-end streaming is too shiny to leave hanging.

PR #35, opened 2015-04-10
Universal output redirection, more exit handling

heavenlyhash:
All output from the CLI can now be shunted to any `io.Writer`.  Tests can use this to check output.  Right now, existing tests route everything to `ioutil.Discard` for lack of any useful assertions.

This cleans up *all* output leaking through test reports, making reading much more pleasant again.

Introduced new errors for user-facing messages, with graceful reporting for and a mechanism for custom exit codes, as well as a polite logging mechanism for unexpected errors.

### DIFF
--- a/repeatr.go
+++ b/repeatr.go
@@ -13,7 +13,7 @@ import (
 
 func main() {
 	try.Do(func() {
-		cli.App.Run(os.Args)
+		cli.Main(os.Args, os.Stderr)
 	}).Catch(cli.Error, func(err *errors.Error) {
 		// Errors marked as valid user-facing issues get a nice
 		// pretty-printed route out, and may include specified exit codes.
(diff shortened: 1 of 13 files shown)


repeatr.go
```diff
@@ -1,11 +1,85 @@
...
+		return "", err
+	}
+	defer logFile.Close()
+	fmt.Fprintf(logFile, "Repeatr error report\n")
+	fmt.Fprintf(logFile, "====================\n")
+	fmt.Fprintf(logFile, "Date: %s\n", time.Now())
+	fmt.Fprintf(logFile, "\n")
+	fmt.Fprintf(logFile, "Full error:\n")
+	fmt.Fprintf(logFile, "-----------\n")
+	fmt.Fprintf(logFile, "%s\n", caught)
```
kofalt: Print the message without the stack?

kofalt:
LGTM for an initial overhaul.
We 'lose' the ability to print to stderr from the CLI, but we weren't doing that anyway, so...

This looks good until we get more aggressive about an output overhaul.

heavenlyhash:
Added a little more terseness for the default user facing message; should be better.

PR #36, opened 2015-04-13
Handle chunk (un)alignment in cbor streamer.

heavenlyhash:
Previously, an incomplete chunk of cbor flushed to disk could reach the cbor decoder, which would then blow up upon hitting the EOF.  Now, there's another layer of proxying writer to make sure the EOF absorption is handled before reaching the decoder.  (This fixes racy panics when the OS flushes more frequently than we explicitly asked.)


kofalt:
LGTM!

PR #38, opened 2015-04-14
Executor failsafe results

heavenlyhash:
Cleanups and fail-safe behavior for some executor failure modes.

- Always close stream muxes
- Set default values for more return codes.  (Zero values were overlaping with valid values (namely, with exit codes), so initializing an unambiguous value is a big deal for maintaining appearances of sanity in error handling paths.  As a specific example, it makes sense for assertions that the exit code is zero-aka-success should fail if the process never launched.)
- Add some more error condition checks in tests that always should've been there.


kofalt:
LGTM

PR #39, opened 2015-04-16
Tar input

heavenlyhash:
Hashing tar inputs!

This implementation of pure golang tar parsing can produce a precise and consistent hash, and tests cleanly on our ubuntu tarball fixture, complete will device nodes, timestamps, uids, etc etc etc... with results exactly matching what we get out of an exec to gnu tar.

### DIFF
--- a/executor/chroot/chroot_executor_test.go
+++ b/executor/chroot/chroot_executor_test.go
@@ -42,6 +42,18 @@ func Test(t *testing.T) {
 				result := e.Start(formula, def.JobID(guid.New()), ioutil.Discard).Wait()
 				So(result.Error, testutil.ShouldBeErrorClass, input.Error)
 			})
+
+			Convey("The job exit code should clearly indicate failure", FailureContinues, func() {
+				formula.Accents = def.Accents{
+					Entrypoint: []string{"echo", "echococo"},
+				}
+				job := e.Start(formula, def.JobID(guid.New()), ioutil.Discard)
+				So(job, ShouldNotBeNil)
+				So(job.Wait().Error, ShouldNotBeNil)
+				// Even though one should clearly also check the error status,
+				//  zero here could be very confusing, so jobs that error before start should be -1.
+				So(job.Wait().ExitCode, ShouldEqual, -1)
+			})
 		}),
 	)
 
--- a/executor/chroot/chroot_executor_test.go
+++ b/executor/chroot/chroot_executor_test.go
@@ -55,6 +55,7 @@ func Test(t *testing.T) {
 					{
 						Type:     "tar",
 						Location: "/",
+						Hash:     "b6nXWuXamKB3TfjdzUSL82Gg1avuvTk0mWQP4wgegscZ_ZzG9GfHDwKXQ9BfCx6v",
 						URI:      filepath.Join(projPath, "assets/ubuntu.tar.gz"),
 					},
 				},
--- a/input/dispatch/dispatch.go
+++ b/input/dispatch/dispatch.go
@@ -5,6 +5,7 @@ import (
 	"polydawn.net/repeatr/input"
 	"polydawn.net/repeatr/input/dir"
 	"polydawn.net/repeatr/input/tar"
+	"polydawn.net/repeatr/input/tar2"
 )
 
 func Get(desire def.Input) input.Input {
@@ -14,6 +15,8 @@ func Get(desire def.Input) input.Input {
 	case "dir":
 		input = dir.New(desire)
 	case "tar":
+		input = tar2.New(desire)
+	case "exec-tar":
 		input = tar.New(desire)
 	default:
 		panic(def.ValidationError.New("No such input %s", desire))
--- a/input/errors.go
+++ b/input/errors.go
@@ -14,6 +14,25 @@ var Error *errors.ErrorClass = errors.NewClass("InputError") // grouping, do not
 var InputHashMismatchError *errors.ErrorClass = Error.NewClass("InputHashMismatchError")
 
 /*
+	Indicates that an error getting the data an input described.
+	(As contrasted with a `TargetFilesystemUnavailableError`, which is
+	an error that comes while trying to unpack the data source onto a
+	local filesystem for use.)
+
+	Examples include a URI that specified a file that doesn't exist, or
+	an IO error later while reading from that source, etc.
+
+*/
+var DataSourceUnavailableError *errors.ErrorClass = Error.NewClass("InputDataSourceUnavailableError")
+
+// Convenience method for wrapping io errors.
+func DataSourceUnavailableIOError(err error) *errors.Error {
+	return DataSourceUnavailableError.Wrap(errors.IOError.Wrap(err)).(*errors.Error)
+}
+
+// REVIEW: should InputHashMismatchError be a subtype of DataSourceUnavailableError?  (leaning towards probably yes)
+
+/*
 	Indicates that the target filesystem (the one given to `Apply`) had some error.
 */
 var TargetFilesystemUnavailableError *errors.ErrorClass = Error.NewClass("InputTargetFilesystemUnavailableError")
--- a/input/errors.go
+++ b/input/errors.go
@@ -7,13 +7,6 @@ import (
 var Error *errors.ErrorClass = errors.NewClass("InputError") // grouping, do not instantiate
 
 /*
-	Indicates that the input failed to obtain data that correctly
-	matches the hash specifying the input.  This may mean there have
-	been data integrity issues in the storage or transport systems involved.
-*/
-var InputHashMismatchError *errors.ErrorClass = Error.NewClass("InputHashMismatchError")
-
-/*
 	Indicates that an error getting the data an input described.
 	(As contrasted with a `TargetFilesystemUnavailableError`, which is
 	an error that comes while trying to unpack the data source onto a
@@ -30,7 +23,14 @@ func DataSourceUnavailableIOError(err error) *errors.Error {
 	return DataSourceUnavailableError.Wrap(errors.IOError.Wrap(err)).(*errors.Error)
 }
 
-// REVIEW: should InputHashMismatchError be a subtype of DataSourceUnavailableError?  (leaning towards probably yes)
+/*
+	Indicates that the input failed to obtain data that correctly
+	matches the hash specifying the input.  This means there have
+	been data integrity issues in the storage or transport systems involved --
+	either the transport is having reliability issues, or, this may be an
+	active attack (i.e. MITM).
+*/
+var InputHashMismatchError *errors.ErrorClass = DataSourceUnavailableError.NewClass("InputHashMismatchError")
 
 /*
 	Indicates that the target filesystem (the one given to `Apply`) had some error.
--- a/input/tar2/compression.go
+++ b/input/tar2/compression.go
@@ -1,0 +1,69 @@
+package tar2
+
+import (
+	"bufio"
+	"bytes"
+	"compress/bzip2"
+	"compress/gzip"
+	"fmt"
+	"io"
+)
+
+type Compression int
+
+const (
+	Uncompressed Compression = iota
+	Bzip2
+	Gzip
+	Xz
+)
+
+func (compression *Compression) Extension() string {
+	switch *compression {
+	case Uncompressed:
+		return "tar"
+	case Bzip2:
+		return "tar.bz2"
+	case Gzip:
+		return "tar.gz"
+	case Xz:
+		return "tar.xz"
+	}
+	return "[unknown]"
+}
+
+func DetectCompression(source []byte) Compression {
+	// Compression detection patterns borrowed from docker/pkg/archive/archive.go, where they also reside under an Apache v2 license
+	for compression, m := range map[Compression][]byte{
+		Bzip2: {0x42, 0x5A, 0x68},
+		Gzip:  {0x1F, 0x8B, 0x08},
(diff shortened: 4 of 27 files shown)


executor/chroot/chroot_executor_test.go
```diff
@@ -42,6 +42,18 @@ func Test(t *testing.T) {
...
+			Convey("The job exit code should clearly indicate failure", FailureContinues, func() {
+				formula.Accents = def.Accents{
+					Entrypoint: []string{"echo", "echococo"},
+				}
+				job := e.Start(formula, def.JobID(guid.New()), ioutil.Discard)
+				So(job, ShouldNotBeNil)
+				So(job.Wait().Error, ShouldNotBeNil)
+				// Even though one should clearly also check the error status,
+				//  zero here could be very confusing, so jobs that error before start should be -1.
+				So(job.Wait().ExitCode, ShouldEqual, -1)
```
kofalt: Does this fail because the validators haven't added a PATH env var & `echo` will not be found?

input/errors.go
```diff
@@ -14,6 +14,25 @@ var Error *errors.ErrorClass = errors.NewClass("InputError") // grouping, do not
...
+
+*/
+var DataSourceUnavailableError *errors.ErrorClass = Error.NewClass("InputDataSourceUnavailableError")
+
+// Convenience method for wrapping io errors.
+func DataSourceUnavailableIOError(err error) *errors.Error {
+	return DataSourceUnavailableError.Wrap(errors.IOError.Wrap(err)).(*errors.Error)
+}
+
+// REVIEW: should InputHashMismatchError be a subtype of DataSourceUnavailableError?  (leaning towards probably yes)
```
kofalt: Sounds like it, yeah. My only reservation is that if a remote input gave something that didn't match, that's rather serious & may want to explicitly draw concern. SSH MITM spitting, etc

executor/chroot/chroot_executor_test.go
heavenlyhash: This is inside the `Convey("Given a rootfs input that errors"` block.

input/tar2/compression.go
```diff
@@ -0,0 +1,68 @@
...
+		return "tar.xz"
+	}
+	return "[unknown]"
+}
+
+func DetectCompression(source []byte) Compression {
+	for compression, m := range map[Compression][]byte{
+		Bzip2: {0x42, 0x5A, 0x68},
+		Gzip:  {0x1F, 0x8B, 0x08},
+		Xz:    {0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00},
```
kofalt: Mind adding a comment regarding where these identifiers came from?

kofalt: Also, any commentary how closely this ended up matching other compression code around, Docker et al?

input/errors.go
heavenlyhash: Mkay, adding subtype.

input/dispatch/dispatch.go
```diff
@@ -14,7 +14,7 @@ func Get(desire def.Input) input.Input {
 	case "dir":
 		input = dir.New(desire)
 	case "tar":
-		input = tar.New(desire)
+		input = tar2.New(desire)
```
kofalt: I think this is a mistake - or rather, incomplete. The package names are fine but we should keep a way to dispatch GNU-wrapping transports until they're **excessively** dead. I want to be able to trivially fall back on these for awhile.

I propose `gnu-tar` for the old one; thoughts?

input/tar2/goconvey.goconvey
```diff
@@ -0,0 +1,2 @@
+#-timeout=30s
+-short
```
kofalt: Are you sure you want this file included? I didn't see tests with a short marker.

input/tar2/tar_input.go
```diff
@@ -0,0 +1,163 @@
...
+
+			// unroll the tar, copying and accumulating data for integrity check
+			bucket := &fshash.MemoryBucket{}
+			walk(tarReader, destinationRoot, bucket, i.hasherFactory)
+
+			// hash whole tree
+			actualTreeHash, _ := fshash.Hash(bucket, i.hasherFactory)
+
+			// verify total integrity
+			expectedTreeHash, err := base64.URLEncoding.DecodeString(i.spec.Hash)
```
kofalt: We might be using this in a few places already, but is it defined anywhere that we're using base64 url-safe encoding of the checksum? Or is that just already well-understood for shaX.

input/dispatch/dispatch.go
heavenlyhash: "exec-tar"?  No way of knowing if whatever may be on the $path is gnu or... what.

input/tar2/tar_input.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"archive/tar"
+	"bytes"
+	"crypto/sha512"
+	"encoding/base64"
+	"hash"
+	"io"
+	"os"
+	"path"
```
kofalt: Should be `path/filepath`

input/tar2/tar_input.go
heavenlyhash: ¯\_(ツ)_/¯  Wildly undocumented and unlikely to be forward compatible.  Definitely casting a side-eye at this whenever saying "this is alpha, and we aren't doing migrations yet".

Also the use of sha384 is just a fiat presumption right now, and ideally we'll investigate having an upgrade path for that before calling anything 1.0.

lib/fshash/bucket.go
```diff
@@ -1,22 +1,59 @@
 package fshash
 
 import (
+	"archive/tar"
+	"path"
```
kofalt: Same `path/filepath` issue here

input/tar2/tar_input.go
```diff
@@ -0,0 +1,163 @@
...
+			panic(input.DataSourceUnavailableError.New("corrupt tar: %s", err))
+		}
+		hdr := fs.Metadata(*thdr)
+		// filter/sanify values:
+		// names must be clean, relative dot-slash prefixed, and dirs slash-suffixed
+		// times should never be go's zero value; replace those with epoch
+		hdr.Name = path.Clean(hdr.Name)
+		if hdr.Name != "." {
+			hdr.Name = "./" + hdr.Name
+		}
```
kofalt: Are you sure this is sufficient to remove all possible escape paths?

From `path/filepath` (which you're not currently using, see above):

> Eliminate .. elements that begin a rooted path: 
> that is, replace "/.." by "/" at the beginning of a path, assuming Separator is '/'.

Wouldn't a non-root-prefixed header name such as `example/../../../etc/passwd` escape?

input/tar2/tar_input.go
heavenlyhash: No, it shouldn't.  It's being used to clean paths fed into hashing.  The canonical form of paths for all internal handling is '`/`' -- Hashes should not be different on platforms with unusual opinions about path formatting.

input/dispatch/dispatch.go
kofalt: `exec-tar` would work. Don't forget to change the name so that it doesn't panic.

input/tar2/tar_input.go
kofalt: Does the tar spec dictate that separators are `/`? 

Point being I'd like a quick comment left somewhere; file-handling code not using `filepath` is a bit surprising, particularly when test code in the same package imports `filepath`.

input/tar2/tar_input.go
heavenlyhash: Oh dear.  [Appears that it would](https://play.golang.org/p/E-D7Z1cVRL).

Well, I'd rather not invoke `Abs` if we can avoid it.  Can we just check for and blacklist a `../` prefix after the `Clean`, and raise that as a corrupt tar?

input/tar2/tar_input.go
heavenlyhash: Spec: don't care, that's what the `path.Clean` is supposed to handle.

This *isn't* file handling code.  The test code using `filepath` *is* file handling code.

Do you think we're going to improve readability by commenting every single use of `path` and `filepath` with a per-line "yes, author knew what was doing" signoff comment?  There are a lot of places that deal with files, and a lot of places that deal with canonicalized data.  That'll generate... a lot of comments.

input/tar2/goconvey.goconvey
heavenlyhash: ... that's odd.  Must have lost it on a rebase.  I'll lose this too, to match, and add it back in separately later.

input/tar2/tar_input.go
kofalt: From a different angle, what's the canonical representation of sha384? Git uses hex, but, well.

input/tar2/tar_input.go
kofalt: Retraction; the `fs.PlaceFile` dictates that 

> No changes are allowed to occur outside of `destBasePath`

input/tar2/tar_input.go
kofalt: I don't think `path.Clean` will handle that at all; none of its steps mention anything about different separators.
But that might be academic since in practice, we only really operate on one architecture.

I was actually thinking about just commenting the *import*, not each *usage*. Would you be amicable to that?

input/tar2/tar_input.go
kofalt: IMO, I would resolve this thread by adding a line to the comment block saying the path is NOT escape-safe at this point & that `fs.PlaceFile` handles escape attempts.

This will guard against us possibly adding code later that is not safe against escape attempts; I want to be really clear which parts of the system are handling safe or unsafe data.

heavenlyhash: Oh, phew.  I knew (well, hoped, obviously) that had to be caught somewhere.

Gonna go double-safe.  It makes sense to immediately reject inputs that look like malice, and label them as such.

kofalt:
Review complete; this looks like excellent work.

Will do a final skim & merge once outstanding threads resolved.

input/tar2/tar_input.go
heavenlyhash: I don't think there is one.  https://github.com/jbenet/multihash is an interesting attempt and maybe we should consider using it.

But for now I still want to explicitly handwave that we're not migration-proofed on this front.  The hard part in this scenario isn't actually including a versioning bit; the semantics require a lot of thought.  (I.e. since changing the hash is the equivalent of a "major version" bump in semver terms, compatibility during a transition is... limited, essentially requiring multiple full "messages" in both relevant formats.  And that... you can see how that gets nontrivial fast.)  We should tackle it someday, but tl;dr I expect it's going to be a very big line-item on its own.

kofalt: Cool link. 

I actually think you're combining two problems into one here: I expect to run into same-hash-type comparison challenges before I expect to manage a full-blown multi-hash cross-compatibility system. Both are inevitable; I assert the later is far larger & further down the road.

Compatibility across hash algorithms would require stuff like telling an output that yeah, actually we need an MD5 for ~reasons~, and stuff like that. Contrast with what I'm looking for, which is merely "have a clearly defined way to represent hashes outside repeatr".

Which is a long-winded way of saying: I'm fine with treating repeatr hashes as opaque utf8 strings for now, and tackling hash representation later & before tackling cross-hash stuff. Multihash looks like a good starting point...

input/tar2/tar_input.go
heavenlyhash: I think it would have trouble being meaningful.  A file may use both packages.  Also, I'm wary of comments in imports.  There's... talk on the golang-dev list about those becoming a mechanical Thing for dependency management.

So I guess I want... a broad sane way to address the boundary between path/filepath without microcomments, and I was roughly hoping one would evolve along package lines that's suuuper clear and high level. But until that becomes something we can do clearly... local bandage comments are maybe better than nothing -.-

kofalt: Sounds good. I'd be comfortable swapping them out once there's something better.

input/tar2/tar_input.go
kofalt: I suppose we can also put and e2e test constructing a tarball that tries to place a file one folder above it's root, then running the input & checking if it escaped. We could have a variety of confirm-jailed tests for stuff like `example/../escapee`

input/tar2/tar_input.go
heavenlyhash: You're right, [`Clean` doesn't handle dirseps](https://play.golang.org/p/-L_Ca9hgTI).

... I guess that's not wrong.  '\' is a valid part of a filename on unix.  Heck, it is on windows too; if you talk to nfts directly without explorer, it'll let you do all sorts of things (or at least, my fuse-ntfs mount on linux is letting me `mkdir \\`, which is a little scary).  So `..\..` might *look* funny, but it is actually a valid filename, and we should accept it.

A build of the CLI components (serving as an uploader, without the containment components) on windows might want to see things differently... but we'll cross that bridge when we come to it, I guess.

input/tar2/goconvey.goconvey
heavenlyhash: Oh, there it is: `testutil.Convey_IfSlowTests` is a thing now.

This is super ugly, I know, and easy to miss: it's one layer nested inside of the IfHaveRoot check.

I want a system for specifying multiple requirements sanely; this ain't it.  :soon:

heavenlyhash:
Comments addressed plus some more top level integration tests appeared for CI.

PR #42, opened 2015-04-30
S3 IO

heavenlyhash:
MVP on using s3 as a data silo.

S3 is a k/v store and not an immediate mapping to a filesystem (despite what the UI may seem to suggest) -- to map to a full filesystem we need more bits for file attributes, etc.  This implementation uses tars as the metaphor for mapping filesystems onto s3.  (Another option would be using something like https://github.com/s3fs-fuse/s3fs-fuse -- this project has its own metaphor for mapping filesystems onto s3.  We decided not to use that for MVP because it's a much more complicated system; with this tarball approach, we know we can localize all network work before and after the job, rather than have interesting issues with random latency on disk mid-job.  http://dsl-wiki.cs.uchicago.edu/index.php/Performance_Comparison:Remote_Usage%2C_NFS%2C_S3-fuse%2C_EBS has some other interesting performance notes; there's no short answers to what's "fastest", apparently.  We can make more/different implementations in the future.)

https://github.com/rlmcpherson/s3gof3r was used to do the data shoveling and aws attribute auth signing stuff; appears to work really well.

Configuration of AWS secret tokens and keys is handled through environment variables for the time being.  We will probably want to expand on this later, building a general secrets management system, for at least two reasons: env vars placed before the entire repeatr system do not have the greatest scope limitations, and also they're effectively a global (what if someone wants to use two different s3 accounts?).  Works for now.

Includes a bunch of enhancements to the flexibility of the IO system shared test specs.  Tests now accept parameters for URIs.

This also demos the use of different URI schemes to switch storage behavior.  The "s3://" scheme will cause the rest of the URI location to be taken literally; the "s3+splay://" scheme will instead enable content-addressable naming of uploads, which means the same location string can be used for as a prefix for as much data as you want, and it will DTRT.  (Presuming this goes well, we should consider using URI schemes in the other transports too; this would also let us drop the current "type" field assertions and have transports check they recognize URI schemes, which makes a lot more sense when we get to configurable/pluginable dispatches for the IO transport systems.)  We've also talked about upgrading data location URIs to more complex structural data than strings can handle, but without any use cases or specific demos on the table yet, well... URIs still work well enough for jdbc after a decade or so, and we're not so different.

Setting up S3/AWS permissions models are *not* handled in this.  Those are arcana that should be managed by your organization, not repeatr.

### DIFF
--- a/input/s3/s3_input.go
+++ b/input/s3/s3_input.go
@@ -1,0 +1,1 @@
+package s3
--- a/input/s3/s3_input.go
+++ b/input/s3/s3_input.go
@@ -1,1 +1,133 @@
 package s3
+
+import (
+	"archive/tar"
+	"bytes"
+	"crypto/sha512"
+	"encoding/base64"
+	"hash"
+	"io"
+	"net/url"
+	"os"
+	"path"
+	"time"
+
+	"github.com/rlmcpherson/s3gof3r"
+	"github.com/spacemonkeygo/errors"
+	"github.com/spacemonkeygo/errors/try"
+	"polydawn.net/repeatr/def"
+	"polydawn.net/repeatr/input"
+	"polydawn.net/repeatr/input/tar2"
+	"polydawn.net/repeatr/lib/fshash"
+)
+
+const Type = "s3"
+
+var _ input.Input = &Input{} // interface assertion
+
+type Input struct {
+	spec          def.Input
+	hasherFactory func() hash.Hash
+}
+
+func New(spec def.Input) input.Input {
+	if spec.Type != Type {
+		panic(errors.ProgrammerError.New("This input implementation supports definitions of type %q, not %q", Type, spec.Type))
+	}
+	return &Input{
+		spec:          spec,
+		hasherFactory: sha512.New384,
+	}
+}
+
+func (i Input) Apply(destinationRoot string) <-chan error {
+	done := make(chan error)
+	go func() {
+		defer close(done)
+		try.Do(func() {
+			// do make a dir for untaring into.
+			// tars may specify permission and time bits for their top dir, but if not, we'll start with sane defaults.
+			err := os.MkdirAll(destinationRoot, 0755)
+			if err != nil {
+				panic(input.TargetFilesystemUnavailableIOError(err))
+			}
+
+			// parse URI
+			u, err := url.Parse(i.spec.URI)
+			if err != nil {
+				panic(input.ConfigError.New("failed to parse URI: %s", err))
+			}
+			bucketName := u.Host
+			storePath := u.Path
+			var splay bool
+			switch u.Scheme {
+			case "s3":
+				splay = false
+			case "s3+splay":
+				splay = true
+			default:
+				panic(input.ConfigError.New("unrecognized scheme: %q", u.Scheme))
+			}
+
+			// load keys from env
+			// TODO someday URIs should grow smart enough to control this in a more general fashion -- but for now, host ENV is actually pretty feasible and plays easily with others.
+			keys, err := s3gof3r.EnvKeys()
+			if err != nil {
+				panic(S3CredentialsMissingError.Wrap(err))
+			}
+
+			// initialize reader from s3!
+			getPath := storePath
+			if splay {
+				getPath = path.Join(storePath, i.spec.Hash)
+			}
+			s3reader := makeS3reader(bucketName, getPath, keys)
+			defer s3reader.Close()
+
+			// prepare decompression as necessary
+			reader, err := tar2.Decompress(s3reader)
+			if err != nil {
+				panic(input.DataSourceUnavailableError.New("could not start decompressing: %s", err))
+			}
+			tarReader := tar.NewReader(reader)
+
+			// unroll the tar, copying and accumulating data for integrity check
+			bucket := &fshash.MemoryBucket{}
+			tar2.Extract(tarReader, destinationRoot, bucket, i.hasherFactory)
+
+			// hash whole tree
+			actualTreeHash, _ := fshash.Hash(bucket, i.hasherFactory)
+
+			// verify total integrity
+			expectedTreeHash, err := base64.URLEncoding.DecodeString(i.spec.Hash)
+			if !bytes.Equal(actualTreeHash, expectedTreeHash) {
+				done <- input.InputHashMismatchError.New("expected hash %q, got %q", i.spec.Hash, base64.URLEncoding.EncodeToString(actualTreeHash))
+			}
+		}).Catch(input.Error, func(err *errors.Error) {
+			done <- err
+		}).CatchAll(func(err error) {
+			// All errors we emit will be under `input.Error`'s type.
+			// Every time we hit this UnknownError path, we should consider it a bug until that error is categorized.
+			done <- input.UnknownError.Wrap(err).(*errors.Error)
+		}).Done()
+	}()
+	return done
+}
+
+var s3Conf = &s3gof3r.Config{
+	Concurrency: 10,
+	PartSize:    20 * 1024 * 1024,
+	NTry:        10,
+	Md5Check:    false,
+	Scheme:      "https",
+	Client:      s3gof3r.ClientWithTimeout(15 * time.Second),
+}
+
+func makeS3reader(bucketName string, path string, keys s3gof3r.Keys) io.ReadCloser {
+	s3 := s3gof3r.New("s3.amazonaws.com", keys)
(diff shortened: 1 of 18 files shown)


input/s3/s3_input.go
```diff
@@ -0,0 +1,133 @@
...
+			bucket := &fshash.MemoryBucket{}
+			tar2.Extract(tarReader, destinationRoot, bucket, i.hasherFactory)
+
+			// hash whole tree
+			actualTreeHash, _ := fshash.Hash(bucket, i.hasherFactory)
+
+			// verify total integrity
+			expectedTreeHash, err := base64.URLEncoding.DecodeString(i.spec.Hash)
+			if !bytes.Equal(actualTreeHash, expectedTreeHash) {
+				done <- input.InputHashMismatchError.New("expected hash %q, got %q", i.spec.Hash, base64.URLEncoding.EncodeToString(actualTreeHash))
```
kofalt: Am I correct in understanding that, to use this implementation, you're required to:

1. Only use s3 assets produced (or scanned) by repeatr's custom sha384 logic
1. Preserve the custom sha384 along with the upload URL, and not the s3-specific hash (md5)

To use this transport?

For example, it sounds like the workflow to use s3 assets right now would be:

1. Download the s3 asset with some other tool (probably the `gof3r` binary, actually)
1. Scan it with repeatr
1. Provide it to a formula, probably as a `dir` input

Both limitations should be fine for a demonstration; I want to make sure I fully understand.

input/s3/s3_input.go
```diff
@@ -0,0 +1,133 @@
...
+	return done
+}
+
+var s3Conf = &s3gof3r.Config{
+	Concurrency: 10,
+	PartSize:    20 * 1024 * 1024,
+	NTry:        10,
+	Md5Check:    false,
+	Scheme:      "https",
+	Client:      s3gof3r.ClientWithTimeout(15 * time.Second),
```
kofalt: Are the values for `Concurrency` and `PartSize` pulled from anywhere, or just 'eh' values? 
Did s3 or s3gof3r have any suggestions on the matter?

If either have opinions on what what works well & why, we may as well keep that paper trail around..

output/s3/s3_output.go
```diff
@@ -0,0 +1,190 @@
...
+	Working with S3 permissions: good luck; be advised that in many configurations,
+	accounts may be able to write data which they then instantly lose permission to
+	manage or even read back; this will result in permission denied errors from this
+	system, and may result in dirty state left behind (because we have to permission
+	to clean it up once we've created this, and cannot discover this until it's
+	too late!).  PRs for improving this behavior deeply welcome; this author
+	is not an S3/AWS/IAM professional.
+
+	Try the docs here: https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html --
+	these describe attaching policies to a bucket that can grant permissions to a user account.
```
kofalt: This is excellent, and I'd like to see either of these added:

1. Parts of the [PR submission comment](https://github.com/polydawn/repeatr/pull/42#issue-72183745) copied into a comment on the s3 input struct, or
1. Just add a link to said comment, until we get our act together documentation-wide

Thus hopefully alleviating "how does this work why no docs."

output/s3/s3_output.go
```diff
@@ -0,0 +1,190 @@
...
+			// this allows us to start uploading before the final hash is determined and relocate it later.
+			// for direct paths, upload into place, because aws already manages atomicity at that scale (and they don't have a rename or copy operation that's free, because uh...?  no time to implement it since 2006, apparently).
+			putPath := storePath
+			if splay {
+				putPath = path.Join(path.Dir(storePath), ".tmp.upload."+path.Base(storePath)+"."+guid.New())
+			}
+			s3writer := makeS3writer(bucketName, putPath, keys)
+
+			// walk, fwrite, hash
+			o.spec.Hash = tar2.Save(s3writer, basePath, o.hasherFactory)
```
kofalt: For s3, what limitations are in the path? Your test mentions `repeatr-test/test-9/rt-splay/heap/`, and I see dots being used in `splay` mode. What I'm ruminating at is that we don't have a way to ask repeatr to gzip and save some 65% space on my jquery build, etc. Potentially, in non-splay mode we could look for an extension, but I'd be unhappy.

This hits our previous discussion about our unfortunate blindness of conflated storage vs transport, and should not be solved in this PR. So just for future reference, how does s3 limit storage URLs?

kofalt:
This looks great! Two limitations I'd like to solve:

1. Omitting the hostname from the URI - arbitrarily many s3 endpoints, [third-party implementations](https://github.com/andrewgaul/s3proxy), etc.
2. Omitting the transport from the URI - s3 scheme is [configurable](http://godoc.org/github.com/rlmcpherson/s3gof3r#Config), don't use SSL on internal proxy, etc

Fixing these two points will be pretty easy without undue effort or going down the rabbit hole. I think we should gun for flexibility & simplicity here by just saying "yeah, we don't know what our final destination for URIs might be, for now we're just going to concatenate".

```
Current:  s3+splay://repeatr-test/test-9/rt-splay/heap/
Proposed: s3+splay://https://s3.amazonaws.com/repeatr-test/test-9/rt-splay/heap/
```

From [net/url](http://golang.org/pkg/net/url/#URL): 
> The general form represented is: 
> scheme://[userinfo@]host/path[?query][#fragment]

And so it becomes trivial to parse the URI twice, in a [safe & reasonable manner](http://play.golang.org/p/CGB_PpiuEx):

```
Temporary repeatr s3 type: s3+splay
Actual s3 address:         https://s3.amazonaws.com/repeatr-test/test-9/rt-splay/heap/
Bucket name:               repeatr-test
Store path:                test-9/rt-splay/heap/
```

Above is an example of doing just that - calling Parse twice. Pinch silly; KISS principle.

input/s3/s3_input.go
heavenlyhash: Correct, AWS's use of md5 is... *cough*.

If you have any existing asset in s3, you can use the non-"splay" type of location to load it... though if you're using repeatr as a coherent system, that's probably going to be an unusual situation to be in.

But yes, we badly need a `repeatr scan` cli command to make it to the tip of the todo heap.

input/s3/s3_input.go
heavenlyhash: Super eh.  I explicitly do *not* have coherent or defensible opinions :P  There's a similar struct embeded in s3gof3r from which most of these are cargo-culted and contains the rousing documentation that "DefaultConfig contains defaults".

kofalt: I assumed as much. We'll keep it until the sun explodes or someone comes waving better cargo-cult.

heavenlyhash: :sun_with_face: 
:camel: 

How about a cargo camel carting a sun cult?

kofalt: EVEN BETTER

output/s3/s3_output.go
heavenlyhash: Moving docs to package (and a separate `doc.go` file, as seems popular when docs get long enough), and adding a ton of stuff...

heavenlyhash: Oh, for inputs?  Yeahhhhh :shifty_eyes:  These'll stop being separate packages when io-reboot lands, and that strange fork in where to look for docs will disappear in a poof of smoke then.

output/s3/s3_output.go
heavenlyhash: I don't think there are any significant limitations in s3 paths.  At least, not that I've noticed yet.

We probably could/should do compression at some point (it's not in our local fs tars either, atm).  I'd say when we do, we're free to use either suffixes to hint it, or rely on the "magic" header bytes (inputs already currently switch on those, regardless of suffixes).  At any rate it shouldn't be a huge pain to migrate in the future, since hashes are on the structural form of the content rather than the raw binary stream; compression doesn't impact them.

heavenlyhash:
These are valid things to worry about, but also at the moment I don't want to go too far down the road of writing code for things we don't have tests for yet.

(Boy, it'd be cool to have an S3 test we can run against an in-memory store in our own sandbox, though.)

Another alternative would put stuffing more parameters like whether or not to use https in query params.  That's what JDBC stuff usually did, iirc.

heavenlyhash:
Docs should stand much improved... I think I'm happy enough to consider merging this (expecting many more PRs to enhance in the future).

output/s3/s3_output.go
kofalt: Works for me!

kofalt:
Okay, that sounds reasonable for an MVP. I'll merge this as-is and we can come back to my comment when it's time to come back & design these properly.

PR #43, opened 2015-04-30
Test requirements system

heavenlyhash:
Tests with a bunch of prereqs were getting clunky to express.  Replaced the prior pattern with https://github.com/smartystreets/goconvey/wiki/Decorating-tests-to-provide-common-logic instead.

Also now generates a short report node to tell you *why* a test group was skipped.

### DIFF
--- a/testutil/requirements.go
+++ b/testutil/requirements.go
@@ -69,14 +69,14 @@ func Requires(items ...interface{}) func(c convey.C) {
 		}
 	}
 	// check requirements
-	var whynot bytes.Buffer
+	var requirementsListing bytes.Buffer
 	var names []string
 	allSat := true
 	for _, req := range requirements {
 		sat := req.Predicate()
 		allSat = allSat && sat
 		names = append(names, req.Name)
-		fmt.Fprintf(&whynot, "requirement %*q: %v\n", widest+2, req.Name, sat)
+		fmt.Fprintf(&requirementsListing, "requirement %*q: %v\n", widest+2, req.Name, sat)
 	}
 	// act
 	if allSat {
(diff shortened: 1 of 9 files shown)


testutil/requirements.go
```diff
@@ -39,3 +42,87 @@ func Convey_IfSlowTests(items ...interface{}) {
...
+	}
+	// check requirements
+	var whynot bytes.Buffer
+	var names []string
+	allSat := true
+	for _, req := range requirements {
+		sat := req.Predicate()
+		allSat = allSat && sat
+		names = append(names, req.Name)
+		fmt.Fprintf(&whynot, "requirement %*q: %v\n", widest+2, req.Name, sat)
```
kofalt: Amusing, but I suspect you'll find `Sprintf` more to your liking :D

testutil/requirements.go
```diff
@@ -39,3 +42,87 @@ func Convey_IfSlowTests(items ...interface{}) {
...
+			case func(c convey.C):
+				action(c)
+			}
+		}
+	} else {
+		title := "Prereqs: " + strings.Join(names, ", ")
+		return func(c convey.C) {
+			convey.Convey(title, nil)
+			c.Println()
+			c.Print(whynot.String())
```
kofalt: Why return all prereqs here, rather than the first one that failed? Is there a way to tell which of the prereqs failed?

testutil/requirements.go
```diff
@@ -39,3 +42,87 @@ func Convey_IfSlowTests(items ...interface{}) {
...
+	action := items[len(items)-1]
+	// examine requirements
+	var widest int
+	for _, req := range requirements {
+		if len(req.Name) > widest {
+			widest = len(req.Name)
+		}
+	}
+	// check requirements
+	var whynot bytes.Buffer
```
kofalt: Name this `requirementsList` or something

testutil/requirements.go
heavenlyhash: Yup, their checks are all accumulated separately.  The "whynot" print appears inside the "prereqs" heading, and that explains each one line by line.

kofalt: Hence the `sat` bit there. Good.

PR #45, opened 2015-05-11
Cli fix

heavenlyhash:
Fix some random issues with the exec-tar output system discovered by running the cli -- the exec tar and the repeatr tar now behave identically.  Switch the default tar implementation to the hashing one (not sure why we didn't do so already; this now matches the pattern already in the inputs dispatch).


kofalt:
LGTM

PR #46, opened 2015-05-14
Add a consistent MkdirAll to lib/fs.

heavenlyhash:
Add a consistent MkdirAll to lib/fs.  Like os.MkdirAll, but flattens out attributes.

This is a feature wanted by the upcoming assembler systems.


kofalt:
LGTM; in the interest of landing these quickly I just went ahead & added a one-liner comment I desired.

PR #50, opened 2015-05-14
Attach hashes structurally to HashMismatchErrors.

heavenlyhash:
Sticking values irretreivably in fmt strings is a poor idea.  Put them into data keys instead, so they can be re-examined later.

Also, standardize the way HashMismatchError are created, so their fmt strings aren't copypasta'd around.


heavenlyhash:
Configuration systems that allow requesting the IO system not raise hash mismatches as errors are coming in io-reboot -- so this will make it possible to still do the check elsewhere and/or any logging and reporting.

PR #51, opened 2015-05-16
Introduce 'Base' dir.

heavenlyhash:
Get all our filesystem content for one run coherently in one place, under one tree, because it makes debugging a heck of a lot easier.

Having the tests of individual packages share a cache is fine and dandy (it's a boon for practical performance, and our cache sanity should have its own unit tests before we hit the heavy lifting code in executor tests); just... make it controllable.

This just converts the default executor config; other change can come as rolling updates to components as we hit them.


kofalt:
The following may also wish to be migrated to use `def.Base`:

* [flak.GetTempDir](https://github.com/polydawn/repeatr/blob/aa133d60123641fa47c38094121f89880bb5d953/lib/flak/assist.go#L25)
* [testutil.WithTmpdir](https://github.com/polydawn/repeatr/blob/aa133d60123641fa47c38094121f89880bb5d953/testutil/tmpdir.go#L29)

Make a decision for each.

On the current tip of `common-work-dir-prefix`, running `goad test` will create:

```
/tmp/repeatr/test/fixtures/
/tmp/repeatr-test/
```

I think the first missed usage I linked above will remove the former directory.

kofalt:
Omitted from the above list is `saveErrorReport` from `repeatr.go`, as I feel that crash reports being dumped outside a potentially-invalid repeatr state folder is fine & dandy.

heavenlyhash:
Updated all the code paths using GetTempDir and removed it.

I can't decide if updating `testutil.WithTmpdir` to be potentially outside of the usual tmpfs seems like a good idea.

heavenlyhash:
Okay, yes, it's a good idea.  Since `./goad` is already putting the base under /tmp in test runs, that point is moot, and after this, we *should* be in a position to check all the files under that base have been cleaned up, if we wanted to add such a test.  So, updating.

kofalt:
Am I correct in thinking that the removal of `input/fixtures/` is due to its wholesale replacement by `testutil/filefixture` functionality?

testutil/filefixture/file_fixtures.go
```diff
@@ -35,6 +35,8 @@ var Beta Fixture = Fixture{"Beta",
 	},
 }.defaults()
 
+var Beta_Hash string = "9GYDihlrhHQRNPV10lms35kogosBekjqJVYzTj0O5H-QJYTU7vf0YAgh3XBWKKBC"
```
kofalt: Do you not have the hash of any other fixtures saved manually?

heavenlyhash:
Yep.

testutil/filefixture/file_fixtures.go
heavenlyhash: Not really, not as part of filefixtures anyway.  They're embedded in a few other tests, though.

And I'm dubious about this one, too.  What we really should have is...?  A map of IO system kinds to a handful of well known hashes?

*Most* of the tests do their own round-trip stuff so that they're not irritatingly fragile if we do perform a protocol update, but it's more an objective than a strictly followed rule.

heavenlyhash: So we could put in an scan round trip inline in the relevant test here, too, it just kinda feels bad.  Maybe the best endgame is to have hashes for fixtures generated automatically at some point for all IO systems?  I'm waiting for personal irritation to boil over enough to embark on that; feel free to tackle.

kofalt: SGTM on all that.

PR #52, opened 2015-05-21
IO reboot

heavenlyhash:
Significant reorganization of IO code, covering three major objectives:

- Move input and output code for a system together, because they're turning out to usually have more to do with their other half than they do with all the other inputs/outputs.
- Make all major interfaces blocking instead of returning channels.
- Reoriented all input data materialization code to *return* a path.  (This is to help make caching possible, and enable zero-copy systems.)
- Prepare transport systems to be more plugin-ready -- the critical interfaces are now composable.

...FOUR major objectives.

- Get parallel assembly working, e.g. so we can start downloading assets A, B, and C in parallel, then toss the rootfs together as quickly as possible after that.

... FIVE... major objectives.

And also:

- Use more typedefs to express things that should already be validated parameters.  Pass those around individually; avoid speaking of the structs in `def`, because those will move to being more the user-facing API (they make sense as a config description, but less so on internals; the number of code paths that need to see e.g. mount location and silo URI at the same time are precisely zilch).
- Make URIs a list.  It's not used much yet (the current API changes mostly shim back to the prior ones, and we'll finish those migrations in a subsequent PR because this one is already way too huge), but it makes A) zero locations make sense for output scans if we just want to get a description of the data without uploading it to storage and B) plural upload locations make sense while doing a single hashing pass.
- Use [Cheney's optional pattern](http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) to get IO systems on a path that allows more complex invocations; some now, and room to expand in the future without wrecking function signatures.

So a lot of objectives.

And on top of all that, this contains working implementations of copy-on-write systems for filesystem assembly (as well as fallbacks to a basic cp-based system), which means if you run a job using a twelve different 30GB assets and it takes a long time to provision... if you run that job *again* it will start in around a hundredth of a second.  This feature applies to both the rootfs (e.g. your ubuntu tarball or whatever) as well as *any other inputs you specify*.

Some names were picked by involvement of a dartboard and might be best considered working titles to be amended in future PRs upon better ideas evolving.

:warning: contains commits from other PRs: rebasing will remove them after #51 lands -- EDIT: done

There's still a lot of to-do's to follow this -- in particular, as mentioned above, the old IO code is still just swaddled by the new interfaces, and should be fully converted -- but the diff size has gotten sufficiently large that it's desirable to land it so we can start working on incremental things again, rather than make this any more unwieldy.


executor/nsinit/nsinit.go
```diff
@@ -131,7 +131,12 @@ func (e *Executor) Execute(f def.Formula, j def.Job, d string, result *def.JobRe
 	cmd.Stderr = errS
 
 	// Prepare filesystem
-	util.ProvisionInputs(f.Inputs, rootfs, journal)
+	transmat := util.DefaultTransmat()
+	util.ProvisionInputs(
+		transmat,
+		util.BestAssembler(),
+		f.Inputs, rootfs, journal,
+	)
```
kofalt: You forgot this line:

`defer assembly.Teardown() // What ever happens: Disassemble filesystem`

Exists in chroot but not nsinit.

kofalt:
This work is great, and has some tasty features written in a compressed timeline. It also (as we've discussed out of band) is very much a work in progress, and my main concern is how landing this enormous changeset (with more to come!) will affect my ability to architecturally re-align the project afterwards.

An incomplete list of qualms:

* "Transmats" seem to hold far too much complexity. Code allegedly about s3 needs to know about storing, scanning, hasing, and more. We should keep that entire category of behaviour as far away from transport-specific code as possible.

* Partly because of the above, a lot of the code is incredibly dense reading. Besides being complex to understand, it looks like a lot of functionality would have to be rewritten to provide app-specific behaviour. 

* I am not convinced multiple URIs are the right level to solve multiple resources sources, and I hearitly disagree with silos being acknowledged at all in the formula. This one isn't too worrying as the formula concept is not done annealing yet.

* s/`io/transmat`/`transport` or something, among other things. Repeatr is already complex, and I suspsect some of these naming + layout choices will seem almost intentionally opaque.

None of this requires a response - I just want to be clear about upcoming refactor targets.

From our out of band conversations, we are heartily agreed that things will get hit with a hammer. Qualms (partially) documented, I am going to merge this, blatantly stating an intention to rework it later.

executor/nsinit/nsinit.go
heavenlyhash: #55

PR #53, opened 2015-05-21
Scan command

heavenlyhash:
CLI upgrades, and a new `repeatr scan` subcommand.

`repeatr scan` invokes the output scan & upload functionality on a local filesystem path.  It produces a json chunk on stdout that can be used directly as configuration for an input.

Some logging that was getting devnull'd has been reconnected (though shoddily -- we need a logging framework, stat).

Fix output of CLIErrors to be more friendly.

 A `demo.sh` script now exists.  This does a slow, semi-interactive showcase of a variety of repeatr features, invoked from bash exactly as a normal user would.

:warning: contains (lots of) commits from other PRs: rebasing will remove them after #52 lands.  EDIT: done.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -4,10 +4,6 @@ import (
 	"io"
 
 	"github.com/codegangsta/cli"
-
-	"polydawn.net/repeatr/def"
-	"polydawn.net/repeatr/executor/dispatch"
-	"polydawn.net/repeatr/scheduler/dispatch"
 )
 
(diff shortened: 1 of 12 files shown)


heavenlyhash:
This should be good to go.

n.b. after this, it's probably high time to add a real logging system.  [log15](https://github.com/inconshreveable/log15) seems to be working nicely in other places I've started using it (good composability on handlers and formatters, etc etc).

PR #57, opened 2015-05-29
io reorg

heavenlyhash:
Reorg of IO code to fall in rank behind the interfaces introduced in #52 , as noted in todos left in that set of changes:

- move prior implementation code to new io package where ins and outs are grouped
- updating tests to latest greatest (i.e. using `testfixture` and shared test specs) along the way, which is generally a net code reduction
- iron out an unnecessary bounce through channels (that's now pushed up to the consumer; we export a synchronous interface)
- make understanding of storage URIs as lists reach into the actual implementations instead of being no-op'd in the wrapper layer
- ...just generally finish carrying things through.

WIP/preview: currently just dir type io, will grow to cover the rest presently.  EDIT: done


heavenlyhash:
Also the currently first commit with the bugfix is gonna get cherry-picked outta here, and *hammered* with more tests because it's *super* irritating that it snuck along this far.

heavenlyhash:
Bugfix extracted to #58.

heavenlyhash:
That's it -- the input and output packages are gone, everything's now migrated to the new interfaces!

Not all migrated systems include full support for multiple storage URIs.  Some future work needed to make those really useful has been discovered.

heavenlyhash:
Delta: +1337, -2083.  Net decrease of 746 lines.  No functionality was dropped (and all additions were leet, apparently!).  Nice.

This is mostly the payoff of universally applying standard compat tests.  There's still a lot of other low hanging fruit for further cleanups.

PR #58, opened 2015-05-31
Fix base dir attribs, mode cannonicalization, and support dev modes

heavenlyhash:
Enhanced test coverage of tar inputs, resulting in several critical fixes:

- Foreign tar streams missing a "./" element now always get deterministic attributes on the base dir, as they should.
- Whitelist modes as fed into hashing more consistently.  (Golang tar has some odd opinions about modes above 0777.)  GNU tar outputs and our own now behave the same.
- Include devmajor and devminor attributes in hash coverage for block and character devices.  Always should have; not doing so has been a security and integrity failure.

There's a flag day on hashes -- there are breaking changes to the hashing algorithm.  Here it is, your "we're pre-alpha" admonition in practice!


input/tar2/tar_input.go
```diff
@@ -75,6 +75,9 @@ func (i Input) Apply(destinationRoot string) <-chan error {
 			bucket := &fshash.MemoryBucket{}
 			Extract(tarReader, destinationRoot, bucket, i.hasherFactory)
 
+			// bucket processing may have created a root node if missing.  if so, we need to apply its props.
+			fs.PlaceFile(destinationRoot, bucket.Root().Metadata, reader)
```
kofalt: A few inquires:

1. Does this comment mean that bucket processing is incorrect / incomplete?
1. Reading this, I initially thought this call would place the entire tar file at the root, clobbering the fs that was just extracted. Does this call only apply metadata because `reader` is at EOF? 
1. If so, could we have a different function that explicitly only applies metadata?

lib/fs/metadata.go
```diff
@@ -125,7 +128,7 @@ func (m Metadata) Marshal(out io.Writer) {
 	enc.EncodeString(magic_UTF8, "t") // type
 	enc.EncodeInt(int64(m.Typeflag))
 	enc.EncodeString(magic_UTF8, "m") // mode -- note this is *not* `os.FileMode`, it's just the perm bits
-	enc.EncodeInt(m.Mode)
+	enc.EncodeInt(m.Mode & 07777)
```
kofalt: Leave a comment explaining why the whitelist. Incidentally, why the leftmost `7` - my assumption was that `1777` is as far as magic bits go; knowing linux I'm probably insufficiently informed.

lib/fs/metadata.go
```diff
@@ -116,6 +116,9 @@ func (m Metadata) Marshal(out io.Writer) {
 	if xattrsLen > 0 {
 		fieldCount++
 	}
+	if m.Typeflag == tar.TypeBlock || m.Typeflag == tar.TypeChar {
+		fieldCount += 2
+	}
```
kofalt: Explain what this does, and why 2?

lib/fshash/hash.go
```diff
@@ -56,6 +56,7 @@ func Hash(bucket Bucket, hasherFactory func() hash.Hash) ([]byte, error) {
 		enc.EncodeMapStart(2) // either way it's header + one of leaves or contenthash
 		enc.EncodeString(magic_UTF8, "m")
 		record.Metadata.Marshal(hasher)
+		//fmt.Printf(":::: %q ->\n\t%#v\n\t%s\n", record.Metadata.Name, record.Metadata, base64.URLEncoding.EncodeToString(hasher.Sum(nil))) // non-cascading
```
kofalt: Do you want this comment in the merge?

lib/fshash/hash.go
```diff
@@ -82,10 +83,15 @@ func Hash(bucket Bucket, hasherFactory func() hash.Hash) ([]byte, error) {
 			// close off the "leaves" array
 			// No map-close necessary because we used a fixed length map.
 			hasher.Write([]byte{0xff}) // should be `codec.CborStreamBreak` but upstream has an export bug :/
+			hash := hasher.Sum(nil)
+			// debug
+			//	if len(strings.Split(record.Metadata.Name, "/")) == 3 {
+			//		fmt.Printf("::: hashing -- %q \t=> %s\n", record.Metadata.Name, base64.URLEncoding.EncodeToString(hash))
+			//	}
```
kofalt: Same here

lib/fs/metadata.go
heavenlyhash: Ugh.  I'm going to claim that the reason this is badly commented is because the responsibility for these bits is too scattered -- it already is commented elsewhere, just not here.  Which is, yes, absolutely still a problem.  "Suffer no jank" seems to apply in force; listening to that smell would have caused this to be noticed a long time ago -.-

I'm becoming increasingly tempted to break the `type Metadata tar.Header` alias and make our own struct, because though the `tar.Header` fields are *mostly* what we want... the edge cases are Extremely Irritating, both in code to handle them, and doubly so to document.  :soon:?

input/tar2/tar_input.go
heavenlyhash: More because `bucket.Root().Metadata` is a dir, but yeah, actually, that reader could/should be nil for explicitness and clarity.

lib/fshash/hash.go
heavenlyhash: i've re-written those enough when "wtf"ing that... yes :I

heavenlyhash:
Docs enhancements applied.

lib/fs/metadata.go
kofalt: Yeah, sounds like a good ticket to file. Within the next few weeks we should trend towards tracking these issues more formally; we might not be there yet, but closer.

kofalt:
LGTM; merge when ready

PR #59, opened 2015-08-02
Make filters specified in configs work.

hdonnay:
Signed-off-by: Hank Donnay <hdonnay@gmail.com>


heavenlyhash:
Awesome!  Thanks!  This is one of the last bits of polish needed to get self-hosted deterministic builds :dancer:

PR #61, opened 2015-09-30
Pretty print output of `run` command

kkroening:
Pretty print output of `run` command.


heavenlyhash:
Usability++, Thanks!

(The failing post-merge test is a regression in the acceptance tests for git repo multihoming that's apparently only highlighted in Travis's environment; I'll merge this now and track that down later because it's totally not from this change.)

PR #63, opened 2015-10-22
Runc executor

heavenlyhash:
Add a new executor plugin, this time backed by containers powered by runc!

Also, there's an asset management and bootstrapping tool wedged in there, which is used to obtain a well known version of a runc executable (e.g. exact same as we used when testing) in a safe integrity-guaranteed way.

### DIFF
--- a/executor/chroot/chroot_executor.go
+++ b/executor/chroot/chroot_executor.go
@@ -123,9 +123,9 @@ func (e *Executor) Execute(f def.Formula, j def.Job, d string, result *def.JobRe
 	// see comments in gosh tests with chroot for information about the odd behavior we're hacking around here;
 	// we're comfortable making this special check here, but not upstreaming it to gosh, because in our context we "know" we're not racing anyone.
 	if externalCwdStat, err := os.Stat(filepath.Join(rootfs, f.Action.Cwd)); err != nil {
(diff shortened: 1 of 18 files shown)


kkroening:
Getting test failures if I don't run as root:
```
  * /home/vagrant/src/repeatr/io/placer/assembler_test.go 
  Line 290: - FSError: IO Error: lchown /tmp/repeatr-test/2015.10.22..04.14.14..1421227418/repeatr-test/089626112/material/alpha: operation not permitted

...

  * /home/vagrant/src/repeatr/io/placer/assembler_test.go 
  Line 309: - FSError: IO Error: lchown /tmp/repeatr-test/2015.10.22..04.14.14..1421227418/repeatr-test/104251186/material/alpha: operation not permitted

...

  * /home/vagrant/src/repeatr/io/placer/assembler_test.go 
  Line 328: - FSError: IO Error: lchown /tmp/repeatr-test/2015.10.22..04.14.14..1421227418/repeatr-test/296492852/material/alpha: operation not permitted
```

heavenlyhash:
Hm, that sucks... those should be fenced in a `testutil.RequiresRoot` requirement.  But it's not from this PR; I'll fix it in another PR shortly, thanks.

kkroening:
LGTM

PR #66, opened 2015-12-22
Google cloud storage

TripleDogDare:
This should enable the use of Google cloud storage.

### DIFF
--- a/io/transmat/gs/gs_stream.go
+++ b/io/transmat/gs/gs_stream.go
@@ -1,86 +1,89 @@
-package s3
+package gs
 
 import (
-	"bytes"
-	"encoding/xml"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
-	"path"
 	"time"
 
-	"github.com/rlmcpherson/s3gof3r"
+	"golang.org/x/oauth2"
+	"google.golang.org/api/storage/v1"
+
 	"polydawn.net/repeatr/io"
 )
 
-var s3Conf = &s3gof3r.Config{
-	Concurrency: 10,
-	PartSize:    20 * 1024 * 1024,
-	NTry:        10,
-	Md5Check:    false,
-	Scheme:      "https",
-	Client:      s3gof3r.ClientWithTimeout(15 * time.Second),
+func httpClient(auth *oauth2.Token) *http.Client {
+	return &http.Client{
+		Transport: &oauth2.Transport{
+			Source: oauth2.StaticTokenSource(auth),
+		},
+	}
 }
 
-func makeS3reader(bucketName string, path string, keys s3gof3r.Keys) io.ReadCloser {
-	s3 := s3gof3r.New("s3.amazonaws.com", keys)
-	w, _, err := s3.Bucket(bucketName).GetReader(path, s3Conf)
+func makeGsObjectService(token *oauth2.Token) *storage.ObjectsService {
+	httpClient := httpClient(token)
+	service, err := storage.New(httpClient)
 	if err != nil {
-		if err2, ok := err.(*s3gof3r.RespError); ok && err2.Code == "NoSuchKey" {
-			panic(integrity.DataDNE.New("not stored here"))
-		} else {
+		panic(GsCredentialsInvalidError.Wrap(err))
+	}
+	objService := storage.NewObjectsService(service)
+	return objService
+}
+
+func makeGsReader(bucketName string, path string, token *oauth2.Token) io.ReadCloser {
+	service := makeGsObjectService(token)
+	response, err := service.Get(bucketName, path).Download()
+	if err != nil {
+		panic(integrity.WarehouseIOError.Wrap(err))
+	}
+	return response.Body
+}
+
+func makeGsWriter(bucketName string, path string, token *oauth2.Token) (io.WriteCloser, <-chan error) {
+	reader, writer := io.Pipe()
+	service := makeGsObjectService(token)
+	object := &storage.Object{Name: path}
+	errCh := make(chan error, 1)
+	go func() {
+		// TODO: multipart or resumable upload using `ResumableMedia`
+		_, err := service.Insert(bucketName, object).Media(reader).Do()
+		if err != nil {
+			errCh <- integrity.WarehouseIOError.Wrap(err)
+		}
+		close(errCh)
+	}()
+	return writer, errCh
+}
+
+func reloc(bucketName, oldPath, newPath string, token *oauth2.Token) {
+	var response *storage.RewriteResponse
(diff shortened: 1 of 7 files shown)


io/transmat/gs/gs_transmat.go
```diff
@@ -0,0 +1,332 @@
...
+func (a dirArena) Teardown() {
+	if err := os.RemoveAll(a.path); err != nil {
+		if e2, ok := err.(*os.PathError); ok && e2.Err == syscall.ENOENT && e2.Path == a.path {
+			return
+		}
+		panic(integrity.TransmatError.New("Failed to tear down arena: %s", err))
+	}
+}
+
+func merge(cs ...<-chan error) <-chan error {
```
TripleDogDare: This can be removed.

io/transmat/gs/gs_stream.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return response.Body
+}
+
+func makeGsWriter(bucketName string, path string, token *oauth2.Token) (io.WriteCloser, <-chan error) {
+	reader, writer := io.Pipe()
+	service := makeGsObjectService(token)
+	object := &storage.Object{Name: path}
+	// TODO: multipart or resumable upload
+	ohshit := make(chan error, 1)
```
heavenlyhash: A more common name for this might be `errCh` :3

io/transmat/gs/gs_token.go
```diff
@@ -0,0 +1,76 @@
...
+	var err error
+	var secrets []byte
+	token := getAccessTokenFromEnv()
+	if token == nil {
+		secrets, err = getServiceAccountSecretsFromEnv()
+		if err == nil {
+			token, err = createAccessTokenFromSecrets(secrets, []string{storage.DevstorageReadWriteScope})
+		}
+	}
+	return token, err
```
heavenlyhash: [Effective Go](https://golang.org/doc/effective_go.html#if) suggests rephrasing `if` statements to return as early as possible to reduce nesting.  I think that would make this function easier to read.

heavenlyhash:
This is awesome!  I'm testing it now and everything looks good.  Google Cloud Storage support is going to be a really nice integration to have.

(I'm also really appreciating the GCS web UI quite a bit compared to how AWS has been treating me.  I might switch the backend behind the repeatr bootstrap assets to GCS after this lands!)

I will have to ask you to do one quick edit though -- the "PR" CI check is red because these commits do not have the DCO sign-off required by the (contribution guide)[https://github.com/polydawn/repeatr/blob/master/CONTRIBUTING.md].  As soon as that's taken care of, I'll be thrilled to merge this :dancer:

io/transmat/gs/gs_token.go
```diff
@@ -0,0 +1,77 @@
...
+	if err == nil {
+		token, err = createAccessTokenFromSecrets(secrets, []string{storage.DevstorageReadWriteScope})
+	}
+	return token, err
+}
+
+func getAccessTokenFromEnv() *oauth2.Token {
+	accessToken := os.Getenv(EnvGsAccessToken)
+	if accessToken != "" {
+		return &oauth2.Token{AccessToken: accessToken}
```
heavenlyhash: Hm.  I wish there was a way to sanity-check the values here to make sure it smells like an oauth token -- It'd be nice to give a "token invalid" error right away without contacting a network service if someone does `export GS_ACCESS_TOKEN="asdflol"` -- but I don't see any way to do that in the library either ([the `Valid()` predicate is... uninspiring](https://github.com/golang/oauth2/blob/2baa8a1b9338cf13d9eeb27696d761155fa480be/token.go#L131)), so okay, I guess we'll do without. :hear_no_evil:

TripleDogDare: This would be nice, but [RFC6749 Section 1.4: Access tokens](https://tools.ietf.org/html/rfc6749#section-1.4) says that access tokens can be opaque to the client and the format is based on the server's requirements. We could implement something based on assumptions about Google's OAuth2 implementation, but ultimately it boils down to "try and see" after eliminating zero values. Anything else will probably be fragile and unsatisfying.

heavenlyhash: Ach.  Consider me well and thoroughly hit with the book.

io/transmat/gs/gs_token.go
```diff
@@ -0,0 +1,71 @@
...
+ Environment variable used to directly pass an OAuth2 access token to repeatr
+*/
+const EnvGsAccessToken = "GS_ACCESS_TOKEN"
+
+/*
+ Environment variable used to pass the path of
+ a Google Developers service account JSON key file
+ which is then used for two-legged OAuth to create an Access Token
+*/
+const EnvGsAccountFile = "GS_SERVICE_ACCOUNT_FILE"
```
heavenlyhash: We should probably add `"GOOGLE_APPLICATION_CREDENTIALS"` as an alias of this as well: https://github.com/golang/oauth2/blob/2baa8a1b9338cf13d9eeb27696d761155fa480be/google/default.go#L62  I like your names here better, TBH, and `GS_ACCESS_TOKEN` in particular is scoped to make sense, but if someone's expecting these defaults to work, might as well.

On the other hand, that `DefaultTokenSource` function also hits `metadata.OnGCE` and that in turn... That's... [rather indiscreet](https://github.com/GoogleCloudPlatform/gcloud-golang/blob/6fdcab499d2c45e2939521c4047855f12a745a1d/compute/metadata/metadata.go#L165) and not the sort of default behavior I want in this project -- phoning home without configuration, potentially adding hundreds of ms of latency in a timeout that can't be configured or disabled, for a feature that's for some people *never used*; all No Thanks.  So, maybe we should copy some of that functionality, and leave the rest out, or bring in a patched version that wraps that in conditionals.

heavenlyhash:
Sans my newly-discovered nit about env var names, this LGTM, CI's green, and I wanna start using this so... There's always time to slather on more config options as needed; Gonna go ahead and merge!

io/transmat/gs/gs_token.go
TripleDogDare: woof :dog2: 
Which environment variables to use and in what contexts can become quite complex. If people download private keys instead of the JSON file then we could load the account name and private keys separately. We could load refresh tokens or we could implement some portion of 3-legged oauth that helps people create tokens using other methods. All of this is extra complexity that we don't need at the moment. I'd like to see some use cases before going full bore into features. And there's a few `TODO`s in there that could be a lot more interesting/useful to knock down such as composite objects.

I agree on avoiding poking extra externals. We aren't trying to support GCE and it should not be wrapped into default behavior.

PR #72, opened 2016-06-18
Change order of $PATH variable

ericevenchick:
If you have Go installed on your system (eg, 1.6), it will use that Go binary. Switched the order so it uses the 1.5 binary.


heavenlyhash:
This has failed on the commit signoff check :P

You can correct this with `git commit --amend -s` and force pushing.

heavenlyhash:
Closing.  It's issue tracker cleanup day.

PR #73, opened 2016-07-19
Allow policy changes in introspective mode.

TripleDogDare:
Adds the option to run `twerk` with different policies.
e.g. `repeatr twerk -P governor`


heavenlyhash:
LGTM.  `cli` package needs to be DRYer in the future but that's... already due for refactor, tbh, and at a bigger scope than this.

PR #74, opened 2016-08-27
Porting to Darwin

ericevenchick:
Because people use OSX apparently.

### DIFF
--- a/lib/fs/emplacements.go
+++ b/lib/fs/emplacements.go
@@ -133,7 +133,7 @@ func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
 	}
 
 	if hdr.Typeflag == tar.TypeSymlink {
-		// need to use LUtimesNano to traverse symlinks
+		// need to use LUtimesNano to avoid traverse symlinks
 		if err := fspatch.LUtimesNano(destPath, hdr.AccessTime, hdr.ModTime); err != nil {
 			ioError(err)
 		}
@@ -147,7 +147,6 @@ func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
 		}
 
 	}
-
 }
 
 // Exposed only because you're probably doing your own trees somehow, and it's
--- a/lib/fs/emplacements.go
+++ b/lib/fs/emplacements.go
@@ -132,16 +132,22 @@ func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
 		}
 	}
 
-	if hdr.Typeflag != tar.TypeSymlink {
+	if hdr.Typeflag == tar.TypeSymlink {
+		// need to use LUtimesNano to traverse symlinks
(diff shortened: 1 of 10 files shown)


lib/fs/emplacements.go
```diff
@@ -132,16 +132,22 @@ func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
 		}
 	}
 
-	if hdr.Typeflag != tar.TypeSymlink {
+	if hdr.Typeflag == tar.TypeSymlink {
+		// need to use LUtimesNano to traverse symlinks
```
heavenlyhash: to *avoid* traversing symlinks

lib/fs/emplacements.go
```diff
@@ -132,16 +132,22 @@ func PlaceFile(destBasePath string, hdr Metadata, body io.Reader) {
...
 		}
-	}
+		if err := fspatch.UtimesNano(destPath, hdr.AccessTime, hdr.ModTime); err != nil {
+			ioError(err)
+		}
 
-	if err := fspatch.LUtimesNano(destPath, hdr.AccessTime, hdr.ModTime); err != nil {
-		ioError(err)
 	}
+
```
heavenlyhash: whitespace police :police_car:

rio/placer/placers_darwin.go
```diff
@@ -0,0 +1,124 @@
+// +build linux
```
heavenlyhash: this is... not doing what you think it is.  in fact it appears to be doing nothing?  the `*_$GOARCH.go` filename appears to have trumped it.

rio/placer/placers_darwin.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+}
+
+var _ rio.Placer = BindPlacer
+
+func BindPlacer(srcPath, destPath string, writable bool, _ bool) rio.Emplacement {
+	panic("BindPlacer unsupported on darwin")
+}
+
+type bindEmplacement struct {
```
heavenlyhash: should be able to drop these types on this plat

rio/placer/placers_darwin.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/spacemonkeygo/errors/try"
+	"go.polydawn.net/repeatr/lib/fs"
+	"go.polydawn.net/repeatr/lib/fspatch"
+	"go.polydawn.net/repeatr/rio"
+)
+
+var _ rio.Placer = CopyingPlacer
+
+func CopyingPlacer(srcBasePath, destBasePath string, _ bool, bareMount bool) rio.Emplacement {
```
heavenlyhash: if this doesn't have any refs mac be hatin on, can we move it to a `placers_allplat.go` file? (or some name like that)

heavenlyhash:
:D

PR #76, opened 2016-09-28
Raceway logging

ericevenchick:


core/cli/serialization.go
```diff
@@ -0,0 +1,78 @@
...
+
+var _ io.Writer = &journalSerializer{}
+
+type ctxPairs []interface{}
+
+func (ctxPairs) MapBySlice() {}
+
+type logItem struct {
+	Level int       `json:"level"`
+	Msg   string    `json:"message"`
```
heavenlyhash: I like the terser `msg` as the serialization key for how often this happens.  Possibly `lvl` along the same reasoning.  Going hog wild, maybe even `t` for time.

core/cli/serialization.go
```diff
@@ -0,0 +1,78 @@
...
+	Data  interface{} `json:"runRecord"`
+	RunID def.RunID   `json:"runID"`
+}
+type journalOutput struct {
+	Data  string    `json:"journal"`
+	RunID def.RunID `json:"runID"`
+}
+type logOutput struct {
+	Data logItem `json:"log"`
+}
```
heavenlyhash:
```go
type row struct {
    RunRecord RunRecord `json:"runRecord,omitEmpty"`
    Journal string `json:"journal,omitEmpty"
    // etc...
}
```

may be an easier way to do this: implicitly a union struct; it wastes some memory, but makes deserialization a lot easier by kind of sidestepping polymorphism.

PR #78, opened 2016-09-28
Git overhaul -- fetch/checkout/submodule splits

heavenlyhash:
Massive overhaul to the git transmat, which should increase efficiency and successful caching.

- Submodules are now cached: if you have a parent project move to a new commit, but keep all the same submodule versions, previously repeatr was oblivious; now, it's a fast 100% cache hit.
- Git data objects are now cached: fetches are incremental when using the same remote repos.
- Slight change: ".git" files in submodule paths no longer leak into the container's sight.  This shouldn't affect you unless you had output configurations exporting filesystems including such files.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule ".gopath/src/github.com/pkg/profile"]
 	path = .gopath/src/github.com/pkg/profile
 	url = https://github.com/pkg/profile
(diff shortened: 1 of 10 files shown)


heavenlyhash:
Whee, re-checked `repeat-thyself.sh` with a build of repeatr before and after this branch -- mission accomplished!  still consistent!  This whole change made with no external impacts.

PR #81, opened 2016-09-29
moved output serialization structs to api

ericevenchick:
Signed-off-by: Eric Evenchick <eric@evenchick.com>


heavenlyhash:
Ideally I think our future might look something like this...

pr

Moving these structs to somewhere under the api package tree like this is a step in the right direction. :+1:

PR #82, opened 2016-10-02
Formula runner API serializables.

heavenlyhash:
`repeatr//api/act` now has some more sub-packages that implement a server and a client for passing the serialized log objects.

The `def.SerializedOutput` type is now known simply as `def.Event`, but serial forms are unchanged.

There's quite a few things still filed as "future work", but the client side should fly right.  (The server proxying side is quite a bit of work to detangle, because the log objects and all the other writers have never before been unified... but what we do emit from stdout should in fact parse through this client.)


api/act/remote/runObserver.go
```diff
@@ -0,0 +1,62 @@
...
+		evt := roc.readOne()
+		if evt == (def.Event{}) {
+			break
+		}
+		stream <- &evt
+	}
+}
+
+func (roc *RunObserverClient) AwaitRunRecord(def.RunID) *def.RunRecord {
+	return nil // TODO
```
ericevenchick: would probably prefer a panic of type ErrUnimplemented or something

api/act/remote/runObserver.go
```diff
@@ -0,0 +1,62 @@
...
+	// It's something you can generate by watching the stream;
+	// it makes no sense to have every stream implementation watch for it itself.
+}
+
+func (roc *RunObserverClient) readOne() def.Event {
+	out := def.Event{}
+
+	err := codec.NewDecoder(roc.Remote, roc.Codec).Decode(&out)
+	meep.TryPlan{
+		{ByVal: io.EOF,
```
ericevenchick: why discard EOFs?

api/act/running.go
```diff
@@ -3,10 +3,69 @@ package act
...
+	*/
+	AwaitRunRecord(def.RunID) *def.RunRecord
+}
+
+type ErrRemotePanic struct {
+	meep.TraitAutodescribing
+	Dump string
+}
+
+type ErrNotMyProblem struct {
```
ericevenchick: I would like docs because this error seems somewhat underspecified

api/act/remote/runObserver.go
heavenlyhash: Now nuked in latest, per the reasoning in that "REVIEW" comment.

api/act/remote/runObserver.go
heavenlyhash: Refactored to actually return a clear state flag.  (Same behavior as before; less weird looking and less abuse of the zero-val of large structs as a sentinel val.)

api/act/running.go
heavenlyhash: Renamed to be less bad.

heavenlyhash:
Added thing to gather dumps on nonsense messages, too.  Pretty happy with this now.

PR #89, opened 2017-03-18
Formula hash func

heavenlyhash:
(Finally!...) Add a `Hash()` function to `def.Formula`.

As the docs on the function describe: this function will return a string that may be expected to converge for formulae which describe identical setups.  Some elements of the structure that are *not* relevant to correctly reproducing the contained task (namely, data storage urls) are explicitly not included under the hash.

This has been a long time in coming: enabling this hash, and its semantics, have been a guiding factor in every part of Repeatr's design.  This hash [*] is effectively a content-addressable description of a computation to be performed, complete with environment.  This can be a primary key in a globally distributable database of *things that can be computed*; a single string usable as an index over arbitrary transformations of any data into any other data.  (Woo.)

<small>[*] or similarly behaving, if slightly different one, after some additional iterations of code</small>

### DIFF
--- a/api/def/base58.go
+++ b/api/def/base58.go
@@ -1,0 +1,92 @@
+package def
+
+// The copy-paste history of this document is long:
+// Originally:
+//   Copyright (c) 2013-2014 Conformal Systems LLC.
+//   Use of this source code is governed by an ISC
+//   license that can be found in the LICENSE file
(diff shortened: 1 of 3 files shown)


tazjin:
Why the strange custom encoding implementation vs. a more standard hex output of some hash function?

PR #91, opened 2017-03-18
fix def: Include conjectured output hashes in formula hash

### LINKED ISSUE #90: Formula hashes should include "conjectured" output hashes
Speaking about [this](https://github.com/polydawn/repeatr/blob/master/api/def/formula.go#L43):

```go
for _, spec := range f2.Outputs {
	spec.Hash = ""
	spec.Warehouses = nil
}
```

(also: argh mutability >.<)

This should only zero output hashes that have `conjecture` set to false.

tazjin:
The hashes of conjectured outputs should be consistent for the same set of
input hashes. It follows that because we include input hashes in the hash,
including output hashes for conjectured outputs should be safe.

This fixes #90

Signed-off-by: Vincent Ambo <tazjin@gmail.com>

### DIFF
--- a/api/def/formula.go
+++ b/api/def/formula.go
@@ -41,8 +41,10 @@ func (f Formula) Hash() string {
 		spec.Warehouses = nil
 	}
 	for _, spec := range f2.Outputs {
-		spec.Hash = ""
 		spec.Warehouses = nil
+		if !spec.Conjecture {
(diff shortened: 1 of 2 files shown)


tazjin:
Conjecture was removed in #97 (which I disagree with) but it breaks this PR.

PR #102, opened 2017-05-25
Change the printed version info

TripleDogDare:
Removes build date because it messes with reproducibility.
Adds a tree hash which should be reproducible and is different
for different dirty builds.
Commit hash is now the full hash instead of the short hash.
Moves the dirty flag to not be attached to the commit hash.
Dirty is now triggered by untracked files.
Adds commit/author dates.

```
repeatr v0.15+dev
git commit: 615f57306c7bfbea934cf264a9230c25775a8115
git dirty: true
git tree hash: 1c55850e869961c0d9636bb2ec5a4a46749e609c
git commit date: Tue, 25 Apr 2017 19:53:43 +0000
git author date: Tue, 25 Apr 2017 19:53:43 +0000
```

Signed-off-by: Calvin Behling <calvin.behling@gmail.com>


TripleDogDare:
Ugh. Okay different git versions and I'm using something old git does not understand. :/

PR #107, opened 2017-07-03
Git2: Reducing git exec dependencies

TripleDogDare:
CI is complaining about the go-git libraries. Not sure what's up with that yet.
This PR needs a little cleaning up yet. But it's far enough along that you can swap out the transport kind in `repeat-thyself.sh` and get correct outputs.

**What does it do?**
Adds a `git2` transmat which uses src-d/go-git library to do git operations. This is for testing purposes at the moment and is expected to be entirely compatible with and eventually replace the regular `git` transmat.

**Goals**
1) Not have dependencies on the host's git installation. This PR will get us _most_ of the way there. The notable exception is that the default local repository transport will still shell out to git.
2) Be able to add features such as configuring a map of ssh keys to use for different repositories.

**Performance**
There should be a _small_ performance gain by fetching and moving less data around and not spinning up new processes. This is largely irrelevant though to our goals here and I didn't bother attempting benchmarks.

**Try it**
Clear the repeatr cache (optional)
`sudo rm -rf /var/lib/repeatr/`
Change `git` to `git2` in `repeat-thyself.sh`
```
	"/task/repeatr/":
		type: "git2"
```
`sudo ./repeat-thyself.sh`
Everything should run fine, with similar output logs and equivalent output hashes.

### DIFF
--- a/core/executor/util/capabilities.go
+++ b/core/executor/util/capabilities.go
@@ -21,6 +21,7 @@ import (
 	"go.polydawn.net/repeatr/rio/transmat/impl/dir"
 	"go.polydawn.net/repeatr/rio/transmat/impl/file"
 	"go.polydawn.net/repeatr/rio/transmat/impl/git"
+	"go.polydawn.net/repeatr/rio/transmat/impl/git2"
 	"go.polydawn.net/repeatr/rio/transmat/impl/gs"
 	"go.polydawn.net/repeatr/rio/transmat/impl/s3"
 	"go.polydawn.net/repeatr/rio/transmat/impl/tar"
@@ -54,6 +55,7 @@ func DefaultTransmat() rio.Transmat {
 		rio.TransmatKind("gs"):   dirCacher,
 		rio.TransmatKind("file"): fileCacher,
 		rio.TransmatKind("git"):  git.New(filepath.Join(workDir, "git")),
+		rio.TransmatKind("git2"): git2.New(filepath.Join(workDir, "git")),
 	})
 	return universalTransmat
 }
--- a/rio/transmat/impl/git2/git_internals.go
+++ b/rio/transmat/impl/git2/git_internals.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"go.polydawn.net/meep"
+	"go.polydawn.net/repeatr/api/def"
 	"go.polydawn.net/repeatr/rio"
 
 	"github.com/inconshreveable/log15"
@@ -88,7 +89,7 @@ func gitCheckout(commit *object.Commit, fs billy.Filesystem) {
 	tree, err := commit.Tree()
 	if err != nil {
 		panic(meep.Meep(
-			&rio.ErrInternal{Msg: "Commit missing tree object"},
+			&def.ErrWareCorrupt{Msg: "Commit missing tree object"},
 			meep.Cause(err),
 		))
 	}
@@ -113,7 +114,7 @@ func listSubmodules(commit *object.Commit, fs billy.Filesystem, commitHash plumb
 	tree, err := commit.Tree()
 	if err != nil {
 		panic(meep.Meep(
-			&rio.ErrInternal{Msg: "Commit missing tree object"},
+			&def.ErrWareCorrupt{Msg: "Commit missing tree object"},
 			meep.Cause(err),
 		))
 	}
@@ -121,7 +122,7 @@ func listSubmodules(commit *object.Commit, fs billy.Filesystem, commitHash plumb
 	cfgModules, err := readGitmodulesFile(fs)
 	if err != nil {
 		panic(meep.Meep(
-			&rio.ErrInternal{Msg: "Unable to read gitmodules file"},
+			&def.ErrWareCorrupt{Msg: "Unable to read gitmodules file"},
 			meep.Cause(err),
 		))
 	}
@@ -130,21 +131,22 @@ func listSubmodules(commit *object.Commit, fs billy.Filesystem, commitHash plumb
 	if cfgModules != nil {
 		for _, submodule := range cfgModules.Submodules {
 			if submodule == nil {
+				// Should protect against incomplete entries
 				panic(meep.Meep(
-					&rio.ErrInternal{Msg: "nil submodule in list"},
+					&def.ErrWareCorrupt{Msg: "nil submodule in list"},
 				))
 			}
 			entry, err := tree.FindEntry(submodule.Path)
 			if err != nil {
 				panic(meep.Meep(
-					&rio.ErrInternal{Msg: "Failed to match submodule entry to tree entry"},
+					&def.ErrWareCorrupt{Msg: "Failed to match submodule entry to tree entry"},
 					meep.Cause(err),
 				))
 			}
 			isSubmodule := entry.Mode == filemode.Submodule
 			if !isSubmodule {
 				panic(meep.Meep(
-					&rio.ErrInternal{Msg: "gitmodule entry is not a submodule"},
+					&def.ErrWareCorrupt{Msg: "gitmodule entry is not a submodule"},
 				))
 			}
 			result[submodule] = entry
@@ -190,7 +192,7 @@ func gitClone(log log15.Logger, remote string, commitHash plumbing.Hash, cacheDi
 		uploadRequest.Wants = []plumbing.Hash{commitHash}
 		if uploadRequest.IsEmpty() {
 			panic(meep.Meep(
-				&rio.ErrInternal{Msg: "Empty upload-pack-request"},
+				&def.ErrWarehouseProblem{Msg: "Empty upload-pack-request"},
 			))
 		}
 		gitFetch(remote, gitStore, uploadRequest)
--- a/rio/transmat/impl/git2/git_internals.go
+++ b/rio/transmat/impl/git2/git_internals.go
@@ -1,0 +1,380 @@
+package git2
+
+import (
+	"io"
+	stdioutil "io/ioutil"
+	"net/url"
+	"os"
+	"path/filepath"
+	"time"
+
+	"go.polydawn.net/meep"
+	"go.polydawn.net/repeatr/rio"
+
+	"github.com/inconshreveable/log15"
+	"gopkg.in/src-d/go-billy.v3"
+	"gopkg.in/src-d/go-billy.v3/osfs"
+	// sdgit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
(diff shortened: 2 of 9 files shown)


heavenlyhash:
Holy :tada: 

This is gonna take me Several Time Units to review but i am So Excited

core/executor/util/capabilities.go
```diff
@@ -54,6 +55,7 @@ func DefaultTransmat() rio.Transmat {
 		rio.TransmatKind("gs"):   dirCacher,
 		rio.TransmatKind("file"): fileCacher,
 		rio.TransmatKind("git"):  git.New(filepath.Join(workDir, "git")),
+		rio.TransmatKind("git2"): git2.New(filepath.Join(workDir, "git")),
```
heavenlyhash: Are we confident enough in this that we should flip this around -- make the new code reached by `"git"` and move the old one to `"git-legacy"`? :dancer:

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "Failed to create git upload-pack session"},
+			meep.Cause(err),
+		))
+	}
+	response, err := session.UploadPack(uploadRequest)
+	if err != nil {
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "git upload-pack-request failed"},
```
heavenlyhash: This might make sense to be `ErrWarehouseProblem` type?

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "Unable to read gitmodules file"},
+			meep.Cause(err),
+		))
+	}
+
+	result := map[*config.Submodule]*object.TreeEntry{}
+	if cfgModules != nil {
+		for _, submodule := range cfgModules.Submodules {
+			if submodule == nil {
```
heavenlyhash: This is guarding against... the `.gitmodules` file having incomplete entries, is that correct?

The `tar` transmat uses `def.ErrWareCorrupt` for things like this; that's probably applicable here too.  (Which I suddenly notice *isn't* in the [list of valid errors](https://github.com/polydawn/repeatr/blob/9be3ceed60b42096455e9ee92ccf41417aa5b105/rio/errors.go#L65-L74) to return from this area -- I'm going to fix that immediately in another branch because it's Clearly Wrong.)

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+				panic(meep.Meep(
+					&rio.ErrInternal{Msg: "nil submodule in list"},
+				))
+			}
+			entry, err := tree.FindEntry(submodule.Path)
+			if err != nil {
+				panic(meep.Meep(
+					&rio.ErrInternal{Msg: "Failed to match submodule entry to tree entry"},
```
heavenlyhash: Same: `def.ErrWareCorrupt`

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+					meep.Cause(err),
+				))
+			}
+			isSubmodule := entry.Mode == filemode.Submodule
+			if !isSubmodule {
+				panic(meep.Meep(
+					&rio.ErrInternal{Msg: "gitmodule entry is not a submodule"},
```
heavenlyhash: Same: `def.ErrWareCorrupt`

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "Commit missing tree object"},
+			meep.Cause(err),
+		))
+	}
+
+	cfgModules, err := readGitmodulesFile(fs)
+	if err != nil {
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "Unable to read gitmodules file"},
```
heavenlyhash: Same: `def.ErrWareCorrupt`.  We can't do our usual degree of checkout if this is missing, so, we must call it corrupt.

Or... we could define our behavior to be doing no submodules in this case.  That could be consistent.  (I think the current git-exec stuff would error in this case.  But I'm more worried about consistency within this implementation than consistency with the vagaries and edge cases of the git-exec system which we'll soon be calling "legacy"...)

So, hm, thoughts?

heavenlyhash: Yeah, `def.ErrWareCorrupt` still seems correct.  You were already treating it as an error, and I agree.  :+1:  Just the type commentary.

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+
+/*
+	Reads the gitmodules file and checks each entry against the commit tree entries.
+	Returns a map of submodule config objects to matching tree entries
+*/
+func listSubmodules(commit *object.Commit, fs billy.Filesystem, commitHash plumbing.Hash) map[*config.Submodule]*object.TreeEntry {
+	tree, err := commit.Tree()
+	if err != nil {
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "Commit missing tree object"},
```
heavenlyhash: Same: `def.ErrWareCorrupt`

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, err
+	}
+	return refs, nil
+}
+
+func gitCheckout(commit *object.Commit, fs billy.Filesystem) {
+	tree, err := commit.Tree()
+	if err != nil {
+		panic(meep.Meep(
+			&rio.ErrInternal{Msg: "Commit missing tree object"},
```
heavenlyhash: Same: `def.ErrWareCorrupt`

rio/transmat/impl/git2/git_transmat.go
```diff
@@ -0,0 +1,205 @@
...
+	transmission: git cannot transmit empty directories.  This can be a major pain.
+	Typical workarounds include creating a ".gitkeep" file in the empty directory.
+	Gitignore files may also inadventantly cause trouble.  Transmat.Materialize
+	will act *consistently*, but it does not overcome these issues in git
+	(doing so would require additional metadata or protocol extensions).
+
+	This transmat is *not* currently well optimized, and should generally be assumed
+	to be re-cloning on all materializations -- specifically, it is not smart
+	enough to recognize requests for different commits and trees from the
+	same repos in order to save reclones.
```
heavenlyhash: Is this comment still true?  (I think it's copypasta from the old git transmat, for which it is also now out of date!  I'll fix that in another branch...)

rio/transmat/impl/git2/git_transmat.go
```diff
@@ -0,0 +1,205 @@
...
+		arena.workDirPath = t.workArea.makeFullCheckoutTempPath(string(dataHash))
+		defer os.RemoveAll(arena.workDirPath)
+
+		log.Info("git: clone starting",
+			"remote", warehouse.url,
+		)
+
+		{
+			started := time.Now()
+			gitClone(log, warehouse.url, CommitId2Hash(dataHash), t.workArea.gitStorageDirs, arena.workDirPath, SubmoduleRecursionDepth)
```
heavenlyhash: Naming nit: I'd prefer to `s/CommitId2Hash/CommitIdToGitHash/`.  The "To" name component is conventional when this kind of conversion comes up elsewhere in the project. :)

rio/transmat/impl/git2/git_transmat.go
```diff
@@ -0,0 +1,205 @@
...
+			log.Info("git: clone complete",
+				"remote", warehouse.url,
+				"elapsed", time.Since(started).Seconds(),
+			)
+		}
+
+		// Since git doesn't convey permission bits, the default value
+		// should be 1000 (consistent with being accessible under the "routine" policy).
+		// Chown/chmod everything as such.
+		if err := fs.Chownr(arena.workDirPath, git_uid, git_gid); err != nil {
```
heavenlyhash: We should be able to get rid of this post-facto loop if we're now able to use our own `PlaceFile` function consistently! :D

rio/transmat/impl/git2/git_warehouse.go
```diff
@@ -0,0 +1,130 @@
...
+	  - So basically, relative paths in git simply Don't Work and are
+	    apparently Not Supported.
+
+	Beyond that, we mostly punt.  It's essentially impossible to validate
+	everything in advance in the same way that git will feel.
+	The area we're *really* minding here is where we've observed
+	`git clone` and `git ls-remote` behaviors to drift apart, because
+	that's exceptionally nasty to deal with or report about clearly.
+*/
+func hammerRelativePaths(coords string) string {
```
heavenlyhash: We should be able to get rid of a looooot of this now that we're almost free of exec land...?

rio/transmat/impl/git2/git_internals.go
```diff
@@ -0,0 +1,380 @@
...
+	defer ioutil.CheckClose(from, &err)
+
+	to, err := fs.OpenFile(f.Name, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, mode.Perm())
+	if err != nil {
+		return
+	}
+
+	defer ioutil.CheckClose(to, &err)
+
+	_, err = io.Copy(to, from)
```
heavenlyhash: There's a decent example of this in the new Rio portages: https://github.com/polydawn/rio/blob/16835742dcbb49e44f786f3f26d5943e99961e68/transmat/tar/tar_packing.go#L91-L111

Major :key: : Make a method that makes a standardized Metadata object, stacks up what metadata git actually gives us on top of that, and then use the result when calling PlaceFile.

That example over yonder reuses a single Metadata object and mutates it to avoid allocations, and I think that's slightly preferable -- but we can ignore that here if you prefer since that's the callback-oriented form this is currently in makes the alloc avoidance near impossible (and doesn't matter terribly since presumably anywhere near here we're IO bound by a huge margin anyway).

core/executor/util/capabilities.go
TripleDogDare: Not at the moment.

rio/transmat/impl/git2/git_internals.go
TripleDogDare: Seems reasonable.

rio/transmat/impl/git2/git_warehouse.go
TripleDogDare: Unsure. Do we have tests for these?

TripleDogDare: We probably should add a few network tests to make sure it all works when not on local fs.

